PR #9, opened 2022-04-14
Add e2e Github action

jveski:
Switches the devenv to use a random resource group name, adds garbage collection tags, and adds a simple GH action to provision a dev env every night. 

Eventually this action will also run some e2e tests.


jveski:
I'm going to slam this one in without a review since it's non-controversial and the project is just getting started.

Leaving this comment to let everyone know that we won't do this once the project is being used in production. :)

PR #19, opened 2022-04-20
Use correct port for upstreams

jveski:
Currently annotated services assume the pods listen on the same port the service listens on. Instead they should use TargetPort - either the name (string) or literal port.


imiller31 (approved):
LGTM

PR #21, opened 2022-04-20
Disable OSM sidecars in placeholder pods

jveski:
The Keyvault CSI placeholder pods don't actually do anything (other than signal to mirror the Keyvault secret) so they don't need OSM sidecars.


jveski:
I'm going to jam this one in without bothering anyone for a review - it's uncontroversial enough and we aren't "productionish" yet. Leaving this note so folks know I won't do this once people are using this project!

PR #22, opened 2022-04-22
Revert #19/#20

jveski:
It turns out OSM doesn't support named ports, so #19 and #20 are irrelevant.

In the process of discovering this, I realized that e2e always passes because every test case uses the same ingress hostname, so all test requests target the first test case's server, which works.

### DIFF
--- a/devenv/main.tf
+++ b/devenv/main.tf
@@ -14,12 +14,8 @@ resource "random_string" "random" {
 
 resource "time_static" "provisiontime" {}
 
-variable "example_ingress_domain" {
+variable "domain" {
   default = "ingress.dev"
-}
-
-variable "example_ingress_host" {
-  default = "test.ingress.dev"
 }
 
 data "azurerm_client_config" "current" {
@@ -48,7 +44,7 @@ resource "azurerm_kubernetes_cluster" "cluster" {
   default_node_pool {
     name       = "default"
     node_count = 2
-    vm_size    = "Standard_DS2_v2"
+    vm_size    = "Standard_DS3_v2"
   }
 
   identity {
@@ -92,19 +88,20 @@ resource "azurerm_key_vault" "keyvault" {
       "Import",
     ]
   }
+}
 
-  access_policy {
-    tenant_id = data.azurerm_client_config.current.tenant_id
-    object_id = data.azurerm_user_assigned_identity.clusteridentity.principal_id
+resource "azurerm_key_vault_access_policy" "allowclusteraccess" {
+  key_vault_id = azurerm_key_vault.keyvault.id
+  tenant_id    = data.azurerm_client_config.current.tenant_id
+  object_id    = data.azurerm_user_assigned_identity.clusteridentity.principal_id
 
-    certificate_permissions = [
-      "Get",
-    ]
+  certificate_permissions = [
+    "Get",
+  ]
 
-    secret_permissions = [
-      "Get",
-    ]
-  }
+  secret_permissions = [
+    "Get",
+  ]
 }
 
 resource "azurerm_key_vault_certificate" "testcert" {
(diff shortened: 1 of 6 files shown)


devenv/main.tf
```diff
@@ -14,14 +14,10 @@ resource "random_string" "random" {
 
 resource "time_static" "provisiontime" {}
 
-variable "example_ingress_domain" {
+variable "domain" {
```
jveski, line 17: shortened for readability

devenv/main.tf
```diff
@@ -14,14 +14,10 @@ resource "random_string" "random" {
...
   default = "ingress.dev"
 }
 
-variable "example_ingress_host" {
```
jveski, line 21: removed because we use a wildcard cert now

devenv/main.tf
```diff
@@ -48,7 +44,7 @@ resource "azurerm_kubernetes_cluster" "cluster" {
   default_node_pool {
     name       = "default"
     node_count = 2
-    vm_size    = "Standard_DS2_v2"
+    vm_size    = "Standard_DS3_v2"
```
jveski, line 47: nodes weren't big enough to schedule all test workloads

devenv/main.tf
```diff
@@ -92,19 +88,20 @@ resource "azurerm_key_vault" "keyvault" {
       "Import",
     ]
   }
+}
 
-  access_policy {
-    tenant_id = data.azurerm_client_config.current.tenant_id
-    object_id = data.azurerm_user_assigned_identity.clusteridentity.principal_id
+resource "azurerm_key_vault_access_policy" "allowclusteraccess" {
```
jveski, line 93: moved the keyvault access policy out of the keyvault resource so it can be provisioned concurrently with the aks cluster

devenv/main.tf
```diff
@@ -148,7 +145,7 @@ resource "azurerm_key_vault_certificate" "testcert" {
       ]
 
       subject_alternative_names {
-        dns_names = ["${var.example_ingress_host}"]
+        dns_names = ["*.${var.domain}"]
```
jveski, line 148: wildcard cert so each testcase can have its own hostname

e2e/e2e_test.go
```diff
@@ -89,50 +86,36 @@ func TestMain(m *testing.M) {
...
 
-	suite.StartTestCase(t).
-		WithResources(
-			fixtures.NewClientDeployment(t, conf.CertHostname, conf.TestNamservers),
-			svr, svc)
-}
-
-// TestBasicServiceWithPortName is identical to TestBasicService but uses a TargetPort that references a port
-// name instead of a literal port int.
-func TestBasicServiceWithPortName(t *testing.T) {
```
jveski, line 124: remove test that is now failing

pkg/controller/service/ingress_reconciler.go
```diff
@@ -71,13 +70,6 @@ func (i *IngressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		serviceAccount = sa
 	}
 
-	upstreamPort := netv1.ServiceBackendPort{}
```
jveski, line 74: no need to honor port names since they don't work anyway

e2e/fixtures/fixtures.go
```diff
@@ -92,8 +92,8 @@ func NewService(app, host, keyvaultURI string, port int32) *corev1.Service {
 		Spec: corev1.ServiceSpec{
 			Ports: []corev1.ServicePort{{
 				Name:       "http",
-				Port:       port,
-				TargetPort: intstr.FromInt(8080),
+				Port:       1234, // anything - we don't use this one
+				TargetPort: intstr.FromInt(int(port)),
```
jveski, line 96: Start using the correct port in e2e - previously we used Port for the upstream connections which is wrong

imiller31 (approved):
LGTM

PR #23, opened 2022-04-25
Enable leader election

jveski:
Allow multiple replicas to run concurrently. Also adds a call to `AddHealthzCheck` since it turns out that is required to register the health probe handler routes.


AYM1607 (approved):
LGTM

PR #25, opened 2022-06-06
Push image to registry for e2e testing

jveski:
Runs the operator on the dev/e2e clusters instead of locally within the gh action or dev workstation.


jveski:
Passing run: https://github.com/Azure/aks-app-routing-operator/actions/runs/2449799976

devenv/main.tf
```diff
@@ -165,41 +173,101 @@ resource "azurerm_role_assignment" "approutingdnszone" {
   principal_id         = data.azurerm_user_assigned_identity.clusteridentity.principal_id
 }
 
-resource "local_sensitive_file" "kubeconfig" {
-  content  = azurerm_kubernetes_cluster.cluster.kube_config_raw
-  filename = "${path.module}/state/kubeconfig"
+resource "azurerm_container_registry" "acr" {
+  name                = "approutingdev${random_string.random.result}a"
```
AYM1607, line 177: Is the "a" there just in case the random string ends in an invalid character?

Makefile
```diff
@@ -9,3 +9,10 @@ clean:
 
 e2e:
 	go test -v --count=1 --tags=e2e ./e2e
+
+push:
+	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator:$(shell date +%s)" > devenv/state/operator-image-tag.txt
+	az acr login -n `cat devenv/state/registry.txt`
+	docker build -t `cat devenv/state/operator-image-tag.txt` .
+	docker push `cat devenv/state/operator-image-tag.txt`
+	kubectl set image --kubeconfig devenv/state/kubeconfig deployments/app-routing-operator operator=`cat devenv/state/operator-image-tag.txt`
```
AYM1607, line 18: What is the advantage of using files here? Any particular reason why we can't use env vars?

e2e/e2eutil/e2eutil.go
```diff
@@ -169,39 +165,3 @@ func (c *Case) ensureNS() {
 		})
 	}()
 }
-
-// SetupRBAC creates and binds a ClusterRole manifest to a given username. Intended to be used
-// by e2e tests that need to limit their own k8s access.
-func SetupRBAC(cs kubernetes.Interface, roleManifestPath, userName string) error {
```
AYM1607, line 175: Is RBAC not set up in code anymore because it we have "kubernetes_cluster_role_binding_v1" in terraform? Or is it unrelated?

devenv/main.tf
jveski: Yes, resource names (at least for some RPs?) can't end in a number

Makefile
jveski: This was just to make troubleshooting easier, but I don't mind changing it back to env vars if you prefer

Makefile
```diff
@@ -9,3 +9,10 @@ clean:
 
 e2e:
 	go test -v --count=1 --tags=e2e ./e2e
+
+push:
+	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator:$(shell date +%s)" > devenv/state/operator-image-tag.txt
```
chihshenghuang, line 14: Where can we get this registry.txt file? create the file locally before running push command?

e2e/e2eutil/e2eutil.go
jveski: Yeah now we use the service account instead of an impersonated identity. We don't care so much about testing with a minimal role since auth is handled differently than we planned in production, so we just use cluster admin now. Omitting details here to avoid leaking details about internal infra :)

Makefile
jveski: yes, it's created by terraform when creating the environment `make dev`

chihshenghuang (approved):

Makefile
AYM1607: files are fine I just wanted to know if there was a specific reason to prefer them over env vars :)

AYM1607 (approved):
LGTM

PR #27, opened 2022-06-29
Enable ingress controller resource cleanup

jveski:
The ingress controller components will now run in kube-system instead of app-routing-system, along with the operator. The managed resources are now owned by the operator so that they will be automatically cleaned up when the operator deployment is deleted.

Back compatibility is maintained by checking if the app-routing-system namespace exists and continuing to use it if so.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -11,7 +11,7 @@ import (
 var Flags = &Config{}
 
 func init() {
-	flag.StringVar(&Flags.NS, "namespace", "app-routing-system", "namespace for managed resources")
+	flag.StringVar(&Flags.NS, "namespace", "app-routing-system", "namespace for managed resources (deprecated: omit flag to use kube-system instead)")
 	flag.StringVar(&Flags.Registry, "registry", "mcr.microsoft.com", "container image registry to use for managed components")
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -28,6 +28,7 @@ func init() {
 	flag.StringVar(&Flags.ServiceAccountTokenPath, "service-account-token-path", "", "optionally override the default token path")
 	flag.StringVar(&Flags.MetricsAddr, "metrics-addr", "0.0.0.0:8081", "address to serve Prometheus metrics on")
 	flag.StringVar(&Flags.ProbeAddr, "probe-addr", "0.0.0.0:8080", "address to serve readiness/liveness probes on")
+	flag.StringVar(&Flags.OperatorDeploy, "operator-deploy", "app-routing-operator", "name of the operator's k8s deployment")
```
AYM1607, line 31: nit: OperatorDeployment :) if you have a reason to use this name instead I'd be interested as to why

AYM1607 (approved):
LGTM

pkg/config/config.go
jveski: Good call - I'll use the full name instead of the shorthand

AYM1607 (approved):

PR #28, opened 2023-01-12
Dynamic provisioning

OliverMKing:
Adds dynamic provisioning to resources needed (ingress controller, ingress class, etc.). This provisioning is based on `ingressClassName`. Essentially, we watch all Ingresses in the cluster and if one has the `ingressClassName` for web app routing then we will provision resources accordingly.

This saves compute power for the user until they are actually using web app routing. This is the first step in our path to supporting multiple ingresses (it doesn't make sense to run resources on a cluster for an ingressclass that won't be used).

TODO: Add e2e test (requires some bigger e2e updates to make them better suited to this)


pkg/controller/informer/ingress_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// add other ingress
+	otherCn := "other.class.com"
+	otherIng := &netv1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{Name: "other"},
+		Spec:       netv1.IngressSpec{IngressClassName: util.StringPtr(otherCn)},
+	}
+	informer.Informer().GetIndexer().Add(otherIng.DeepCopyObject()) // use a deep copy because we update later
+
+	// prove that informer by classname returns all ingresses with a class
```
davidgamero, line 48: nit: probably a different word that "prove" here

pkg/controller/keyvault/ingress_secret_provider_class_test.go
```diff
@@ -188,11 +190,13 @@ func TestIngressSecretProviderClassReconcilerBuildSPCInvalidURLs(t *testing.T) {
 
 	t.Run("url with control character", func(t *testing.T) {
 		ing := ing.DeepCopy()
-		ing.Annotations = map[string]string{"kubernetes.azure.com/tls-cert-keyvault-uri": string([]byte{0x7f})}
+		cc := string([]byte{0x7f})
```
davidgamero, line 193: can we put this character in a constant so the intent is more readable?

pkg/controller/ingress/ingress_controller_reconciler_test.go
```diff
@@ -40,21 +56,60 @@ func TestIngressControllerReconcilerIntegration(t *testing.T) {
...
+		logger:      logr.Discard(),
+		ingInformer: ingInformer,
+		className:   ingressCn,
 	}
 	require.NoError(t, i.tick(context.Background()))
 
-	// Prove the resource exists
+	// Prove the resource doesn't exist yet
 	actual := &corev1.Namespace{}
+	require.True(t, errors.IsNotFound(c.Get(context.Background(), client.ObjectKeyFromObject(obj), actual)), "expected not found error")
```
davidgamero, line 71: can we abstract the expect to exist/ not exist into functions to improve readability instead of having many copies of the same long line with the same comment

PR #30, opened 2023-02-02
Adding Support for Private DNS Zones

jaiveerk:
PR to deploy modified external DNS resources for private DNS zones, plus testing suite that supports private clusters/private DNS zones.

Annotation changes came from this tutorial from ExternalDNS: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/azure-private-dns.md


.github/workflows/e2e-private-cluster.yaml
```diff
@@ -0,0 +1,39 @@
...
+    branches: [main]
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
OliverMKing, line 16: checkout version is outdated. Bump the versions on these

.idea/workspace.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
OliverMKing: this file shouldn't be committed

OliverMKing: you could add it to the gitignore

pkg/config/config.go
```diff
@@ -21,6 +21,7 @@ func init() {
 	flag.StringVar(&Flags.DNSZoneSub, "dns-zone-subscription", "", "subscription ID of the Azure DNS Zone (optional)")
 	flag.StringVar(&Flags.DNSZoneDomain, "dns-zone-domain", "", "domain hostname of the Azure DNS Zone (optional)")
 	flag.StringVar(&Flags.DNSRecordID, "dns-record-id", "aks-app-routing-operator", "string that uniquely identifies DNS records managed by this cluster (optional)")
+	flag.BoolVar(&Flags.DNSZonePrivate, "dns-zone-private", false, "if the user's DNS zone is private")
```
OliverMKing, line 24: Can we detect if a user's dns zone is private instead of adding a flag here? 

Are the rp code changes ready to support this flag ready?

devenv/private-cluster.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
OliverMKing, line 1: It seems incorrect to test this way here. We should be relying on the exact k8s manifests that rp uses to guarantee it, not creating arbitrary manifests that are meaningless in the context of the real application.

.github/workflows/e2e-private-cluster.yaml
```diff
@@ -0,0 +1,39 @@
+name: Run e2e private cluster
+on:
+  workflow_dispatch:
```
OliverMKing, line 3: do we want to set this up on pr? We can make it so a member of our team has to be the one to run

devenv/scripts/deploy_private_e2e_tester.sh
```diff
@@ -0,0 +1,41 @@
...
+
+  sleep 5
+
+done
+
+echo "exited loop with status $STATUSNOQUOTES"
+
+FINALSTATUS=$(echo $RESULT | jq ".exitCode")
+
+echo "Test finished, echoing test pod logs..."
```
OliverMKing, line 31: why are we using this script instead of using go tests like the other e2e tests?

Makefile
```diff
@@ -1,18 +1,37 @@
...
 	go test -v --count=1 --tags=e2e ./e2e
 
-push:
+e2e-private-cluster:
+	# grab the image name and tag - that first gets determined by where and how we push it - so first need to build and push it - then export that name and registry and tag into the YAML
+	az acr login -n `cat devenv/state/registry.txt`
+	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator-e2e:$(shell date +%s)" > devenv/state/e2e-image-tag.txt
+	docker build --platform=linux/amd64 -t `cat devenv/state/e2e-image-tag.txt` -f e2e/Dockerfile .
+	docker push `cat devenv/state/e2e-image-tag.txt`
+	cd devenv && /bin/sh scripts/deploy_private_e2e_tester.sh
```
davidgamero, line 24: switching to /usr/bin/env instead of /bin/sh would make it more compatible with other shells users could have when develping

pkg/manifests/manifests.go
```diff
@@ -227,6 +244,42 @@ func newIngressControllerService(conf *config.Config) *corev1.Service {
 	}
 }
 
+func newIngressControllerServicePrivateDNS(conf *config.Config) *corev1.Service {
```
davidgamero, line 247: extracting several of the magic strings from the Private and Normal version of this function will reduce opportunities for drift

davidgamero (dismissed):

.idea/workspace.xml
jaiveerk: hm I've tried adding it multiple times with git rm, not sure why it's being so persistent. Got it removed earlier but it keeps coming back. Will look into this.

pkg/manifests/manifests.go
```diff
@@ -227,6 +244,42 @@ func newIngressControllerService(conf *config.Config) *corev1.Service {
...
+	return &corev1.Service{
```
OliverMKing, line 248: instead of creating a new object with 90% of it being the same as `newIngressControllerService`, just modify the object that is returned with that function. It looks like you're just adding annotations.

OliverMKing (changes requested):
Left some comments

devenv/private-cluster.yaml
jaiveerk: that's a good point - I just took this from the existing e2e framework. I imagine it's a copy of what RP uses anyway, but I'll look to see what I can do to synchronize the two.

.github/workflows/e2e-private-cluster.yaml
jaiveerk: I tried it but ran into the same issue where github secrets can't be accessed for PRs across forks, so az login fails on PRs

devenv/scripts/deploy_private_e2e_tester.sh
jaiveerk: this uses the same e2e test that is used for public clusters. This script simply runs that same go code on the private cluster by deploying a pod on the private cluster that then calls make e2e.
We can't do it via go code because we can't access private clusters via a Kubernetes client on a machine outside the private cluster's VM.
Plus, this way, we have one suite that runs for both private and public clusters, so if changes are made/tests are added to this one suite, those changes propagate to both kinds of clusters.

pkg/manifests/manifests.go
jaiveerk: yep! just adding annotations. wasn't sure which would be clearer, but can definitely switch to just modifying fields in the existing object if you think that makes more sense/is more readable.

pkg/manifests/manifests.go
jaiveerk: good call!

devenv/private-cluster.yaml
jaiveerk: I think the smartest way to go about this will just be to use some sort of embedded file system that we can import from RP, but given how many other changes we're planning for this add-on, I'm gonna regard that as out of scope and a test enhancement that we can add later as we start changing stuff up on the testing side. Just want to have SOME tests at the moment.

pkg/manifests/manifests.go
```diff
@@ -24,15 +25,62 @@ import (
...
-
-	externalDNSName = "external-dns"
+	IngressPodLabels      = map[string]string{appKeySelector: IngressControllerName}
+
+	externalDNSName         = "external-dns"
+	azurePrivateDNSProvider = "azure-private-dns"
+	ingressSource           = "ingress"
+
+	topLevelLabels             = map[string]string{"app.kubernetes.io/managed-by": "aks-app-routing-operator"}
+	nginxIngressControllerArgs = []string{
```
OliverMKing, line 69: this is another example of something that shouldn't be moved to the top. This only has context inside of the nginxIngressControllerDeployment. Moving it here provides no benefit and just makes it way harder to read the code

OliverMKing (changes requested):

pkg/config/config.go
```diff
@@ -39,6 +40,7 @@ type Config struct {
 	MSIClientID, TenantID                             string
 	Cloud, Location                                   string
 	DNSZoneRG, DNSZoneSub, DNSZoneDomain, DNSRecordID string
+	DNSZonePrivate                                    bool
```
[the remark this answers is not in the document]
jaiveerk: the check happens in RP when users link the DNS zone to their add-on. We already perform resource parsing in RP at that step (to extract information like the DNS zone domain and subscription ID) so it makes the most sense to also check whether or not it's private at that step.

pkg/manifests/manifests.go
```diff
@@ -24,15 +25,62 @@ import (
 	"github.com/Azure/aks-app-routing-operator/pkg/util"
 )
 
-const IngressClass = "webapprouting.kubernetes.azure.com"
+const (
+	IngressClass                = "webapprouting.kubernetes.azure.com"
+	appsV1ApiVersion            = "apps/v1"
+	deploymentKind              = "Deployment"
+	nsKind                      = "Namespace"
```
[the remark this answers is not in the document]
jaiveerk: I'll re-introduce some of the const values as regular strings

pkg/manifests/manifests.go
jaiveerk: makes sense - esp given that a good chunk of this will be ripped out anyway, I'll just reintroduce it as regular values to minimize diff

pkg/config/config.go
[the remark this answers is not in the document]
jaiveerk: documenting our discussion here - agreed that resource parsing can happen on the add-on side, and we don't need to fill everything in via individual flags. can instead parse a lot of that from resource ID that we can feed in as a flag. Refactoring entire flag system will be a pretty big PR here and on RP side that is out of scope for this feature, but will jump on that in a new set of PRs following this one.

devenv/scripts/deploy_private_e2e_tester.sh
```diff
@@ -0,0 +1,41 @@
...
+az aks command invoke --resource-group $CLUSTER_RESOURCE_GROUP --name $CLUSTER_NAME --command "kubectl delete jobs app-routing-operator-e2e -n kube-system"
+
+set -e
+az aks command invoke --resource-group $CLUSTER_RESOURCE_GROUP --name $CLUSTER_NAME --command "kubectl apply -f e2e-tester-formatted.yaml" --file state/e2e-tester-formatted.yaml
+
+RESULT="null"
+while [ "$RESULT" = "null" ]
+do
+  STATUS=$(az aks command invoke --resource-group $CLUSTER_RESOURCE_GROUP --name $CLUSTER_NAME --command 'kubectl get pods --selector=app=app-routing-operator-e2e --output="jsonpath={.items[0].status.containerStatuses[?(@.name==\"tester\")].state}" -n kube-system' -o json | jq ".logs" | tr -d '\\')
+  STATUSNOQUOTES=${STATUS#"\""}
```
davidgamero, line 17: commenting the prefix/suffix removal would make this clearer for anyone who isn't as famiar with bash

devenv/scripts/deploy_private_e2e_tester.sh
```diff
@@ -0,0 +1,41 @@
...
+
+echo "Test finished, echoing test pod logs..."
+az aks command invoke --resource-group $CLUSTER_RESOURCE_GROUP --name $CLUSTER_NAME --command 'kubectl logs -l app=app-routing-operator-e2e -n kube-system'
+
+if [ $FINALSTATUS != "0" ]
+then
+  echo "TEST FAILED"
+fi
+
+
```
davidgamero, line 39: nit: several blank lines here

devenv/main.tf
```diff
@@ -250,25 +317,48 @@ resource "kubernetes_cluster_role_binding_v1" "defaultadmin" {
...
 resource "local_sensitive_file" "kubeconfig" {
-  content  = azurerm_kubernetes_cluster.cluster.kube_config_raw
+  content  = var.private-dns ? azurerm_kubernetes_cluster.cluster-private[0].kube_config_raw : azurerm_kubernetes_cluster.cluster[0].kube_config_raw
   filename = "${path.module}/state/kubeconfig"
 }
 
 resource "local_file" "e2econf" {
   content = jsonencode({
-    TestNamservers    = azurerm_dns_zone.dnszone.name_servers
+    TestNamservers    = azurerm_dns_zone.dnszone[0].name_servers
```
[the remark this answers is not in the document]
jaiveerk: yep good catch!

devenv/scripts/push_addon_image.sh
```diff
@@ -0,0 +1,9 @@
+# Get cluster information for az aks command invoke
+CLUSTER_RESOURCE_GROUP=$(cat state/cluster-info.json | jq '.ClusterResourceGroup' | tr -d '"')
+CLUSTER_NAME=$(cat state/cluster-info.json | jq '.ClusterName' | tr -d '"')
```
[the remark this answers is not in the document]
jaiveerk: good idea! this will definitely make debugging easier

davidgamero (approved):
lgtm

pkg/manifests/manifests.go
```diff
@@ -69,13 +97,31 @@ func IngressControllerResources(conf *config.Config, self *appsv1.Deployment) []
 	return objs
 }
 
+func getExternalDNSDeployment(conf *config.Config, dnsCMHash string) *appsv1.Deployment {
```
jveski, line 100: I'd prefer to strategically set particular flags in the existing DNS deployment rather than forking the args entirely here.

pkg/manifests/manifests.go
```diff
@@ -24,15 +25,42 @@ import (
 	"github.com/Azure/aks-app-routing-operator/pkg/util"
 )
 
-const IngressClass = "webapprouting.kubernetes.azure.com"
+const (
+	IngressClass                = "webapprouting.kubernetes.azure.com"
+	deploymentKind              = "Deployment"
```
jveski, line 30: nit: I personally find it more expressive to keep the kinds in struct string literals since they're commonly seen as literals in yaml manifests. Wouldn't block on it so do what you will. :)

Dig the other consts though - no more digging through the manifest to find the image tag.

pkg/manifests/manifests.go
```diff
@@ -189,17 +233,30 @@ func newIngressControllerClusterRoleBinding(conf *config.Config) *rbacv1.Cluster
...
 		Subjects: []rbacv1.Subject{{
-			Kind:      "ServiceAccount",
+			Kind:      serviceAccountKind,
 			Name:      IngressControllerName,
 			Namespace: conf.NS,
 		}},
 	}
 }
 
+func getIngressControllerService(conf *config.Config) *corev1.Service {
```
jveski, line 243: Would prefer to merge this logic into `newIngressControllerService` to avoid adding another "hop" for future readers

pkg/manifests/manifests.go
```diff
@@ -189,17 +233,30 @@ func newIngressControllerClusterRoleBinding(conf *config.Config) *rbacv1.Cluster
...
+	serviceObj := newIngressControllerService(conf)
+	if conf.DNSZoneDomain != "" && conf.DNSZonePrivate {
+		annotations := map[string]string{
+			"service.beta.kubernetes.io/azure-load-balancer-internal": "true",
+			"external-dns.alpha.kubernetes.io/hostname":               fmt.Sprintf("loadbalancer.%s", conf.DNSZoneDomain),
+			"external-dns.alpha.kubernetes.io/internal-hostname":      fmt.Sprintf("clusterip.%s", conf.DNSZoneDomain),
+		}
+		serviceObj.ObjectMeta.Annotations = annotations
```
jveski, line 251: Update specific keys instead of replacing the entire map pls

e2e/e2e_test.go
```diff
@@ -26,7 +27,7 @@ var (
 )
 
 type testConfig struct {
-	TestNamservers            []string
+	TestNameservers           []string
```
jveski, line 30: ha thanks for correcting 🙃

devenv/scripts/deploy_addon_private_cluster.sh
```diff
@@ -0,0 +1,18 @@
+# Load variables from tfstate
+export CLUSTER_CLIENT_ID=$(cat terraform.tfstate | jq '.resources[] | select((.name=="clusteridentity") and .type=="azurerm_user_assigned_identity") | .instances[] | .attributes.client_id')
```
jveski, line 2: Can we manage a local env file with Terraform that contains the exports? Best to avoid parsing the tfstate since it isn't exactly a stable versioned api

e2e/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.19
```
jveski, line 1: What do you think about running the non-private tests containerized like this as well? Even though it might be a bit more brittle in general, seems like standardizing on one approach for private/non-private is a good enough justification.

devenv/scripts/make_e2e.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+make e2e
```
jveski, line 2: is this file needed?

devenv/scripts/push_addon_image.sh
```diff
@@ -0,0 +1,25 @@
+# Get cluster information for az aks command invoke
```
jveski, line 1: should this script be named `push_image`? From this project's perspective it isn't really an addon

jveski (changes requested):

e2e/Dockerfile
OliverMKing: I think this is the right way as well. It makes it so the point of these e2e tests are solely to test the WAR code while rp tests tests ensure the code there.

devenv/scripts/make_e2e.sh
jaiveerk: I couldn't get the docker image to run the command without calling it from within a shell script. I can try to make it work without it if you'd like, but I spent a while and kept getting a weird "exec" error

pkg/manifests/manifests.go
jaiveerk: I'm not sure what you mean. Do you mean we should have this check (and resulting args) be determined within the newExternalDNSDeployment function? Or do you mean something along the lines of 
- assign initial args
- check if private
- modify individual entries/add necessary entries if so

pkg/manifests/manifests.go
jaiveerk: we're changing this a good deal with @OliverMKing 's next PR anyway, but I've personally run into issues when my string for the kind of resource wasn't exactly what it needs to be, so I wanted to keep a few of them as consts.

devenv/scripts/deploy_addon_private_cluster.sh
jaiveerk: do you mean we should be writing those values to an external file/reading them from there rather than from the tfstate?

e2e/Dockerfile
jaiveerk: I don't see why not! Can definitely look into doing this

devenv/scripts/push_addon_image.sh
jaiveerk: fair point!

devenv/main.tf
```diff
@@ -7,10 +7,10 @@ provider "azurerm" {
 }
 
 provider "kubernetes" {
-  host                   = azurerm_kubernetes_cluster.cluster.kube_config.0.host
-  client_certificate     = base64decode(azurerm_kubernetes_cluster.cluster.kube_config.0.client_certificate)
-  client_key             = base64decode(azurerm_kubernetes_cluster.cluster.kube_config.0.client_key)
-  cluster_ca_certificate = base64decode(azurerm_kubernetes_cluster.cluster.kube_config.0.cluster_ca_certificate)
+  host                   = var.private-dns ? azurerm_kubernetes_cluster.cluster-private[0].kube_config.0.host : azurerm_kubernetes_cluster.cluster[0].kube_config.0.host
```
bfoley13, line 10: With more additions to config and complex feature I think this wont be viable. I think it would be better to have separation in the tf files used for specific tests/dev than clump it all into one place

devenv/scripts/push_addon_image.sh
bfoley13, line 1: Do you think we could split this PR into one thats changing the current E2E testing and then a follow up on enabling private clusters?

bfoley13 (changes requested):

devenv/main.tf
jaiveerk: with the number of repeated resources across both test types I wasn't sure, but are you basically saying we should have one TF file for private clusters and one for public?

devenv/scripts/push_addon_image.sh
jaiveerk: not exactly. combining both regular and private zone e2e testing relies on making the public suite use specific functionality (namely, the `az aks command invoke` command for deploying/monitoring everything). I could in theory try merging one PR for public stuff and then just rewrite whatever needs to be rewritten to make it compatible with private, but that doesn't seem like the smartest way to move forward imo.

pkg/manifests/manifests.go
jaiveerk: prior to my transformation the serviceObj doesn't have a map at serviceObj.ObjectMeta.Annotations - there's only a name, namespace, and labels

pkg/manifests/manifests.go
jveski: > Do you mean we should have this check (and resulting args) be determined within the newExternalDNSDeployment function?

Yes. 🙃

pkg/manifests/manifests.go
jaiveerk: following discussion, I will add a blank map in the original annotations position.

pkg/manifests/manifests.go
```diff
@@ -497,6 +474,19 @@ func newExternalDNSDeployment(conf *config.Config, configMapHash string) *appsv1
 			},
 		},
 	}
+	if conf.DNSZonePrivate {
```
jveski, line 477: Can we DRY out the flags a bit?

Something like this:


```go
// Only set shared values in the struct
						Args: []string{
							"--source=ingress",
							"--interval=3m0s",
							"--txt-owner-id=" + conf.DNSRecordID,
							"--domain-filter=" + conf.DNSZoneDomain,
						},

// Then append relevant flags using logic
var additionalArgs []string
if conf.DNSZonePrivate {
		additionalArgs = []string{
			fmt.Sprintf("--provider=%s", azurePrivateDNSProvider),
			fmt.Sprintf("--azure-subscription-id=%s", conf.DNSZoneSub),
		}
} else {
		additionalArgs = []string{
			"--provider=azure",
		}
}
deploymentObj.Spec.Template.Spec.Containers[0].Args = append(deploymentObj.Spec.Template.Spec.Containers[0].Args, additionalArgs...)
```

pkg/manifests/manifests.go
```diff
@@ -282,6 +252,13 @@ func newIngressControllerService(conf *config.Config) *corev1.Service {
 			},
 		},
 	}
+	if conf.DNSZoneDomain != "" && conf.DNSZonePrivate {
```
jveski, line 255: 👍

devenv/main.tf
```diff
@@ -354,11 +353,28 @@ resource "local_file" "e2econfprivatedns" {
   count = var.private-dns ? 1 : 0
 }
 
-resource "local_file" "private_cluster_info" {
+resource "local_file" "registryconf" {
+  content  = azurerm_container_registry.acr.login_server
+  filename = "${path.module}/state/registry.txt"
+}
+
+resource "local_file" "cluster_info" {
```
jveski, line 361: Do you think it would make sense to fold the values of the two new local_file resources into the old e2econf?

I can see some value in keeping unrelated values in separate files, but at first glance there doesn't seem to be a conceptual difference between the various local files.

jaiveerk: there's no conceptual difference but it has more to do with where the information is needed, namely that the e2e conf gets JSON-marshalled into a struct so I wanted to keep that to the values that are used within the testing suite itself

pkg/manifests/manifests.go
jaiveerk: yep that totally makes sense!

devenv/main.tf
jveski: That makes sense. I would personally jam them all into one file so future consumers of the values don't need to worry about which file to read. But I'm cool either way

jaiveerk: plus I think the conceptual separation makes sense given that the cluster_info isn't used by the test suite but rather the scripts that deploy the test, and that the other local file is just used by private tests (pushing a commit soon that separates the tf files though and that'll demonstrate that last piece pretty clearly)

Makefile
```diff
@@ -1,18 +1,41 @@
-.PHONY: dev clean e2e push
+.PHONY: clean dev dev-private-cluster update-image-on-deployment push-tester-image deploy-e2e run-e2e
+
+# keep separate for simultaneous public/private dev without need for resource recreation
+clean-public:
```
OliverMKing, line 4: would be good to just have a `clean` that does both as well

Makefile
```diff
@@ -1,18 +1,41 @@
...
+	docker build --platform=linux/amd64 -t `cat devenv/state/e2e-image-tag.txt` -f e2e/Dockerfile .
+	docker push `cat devenv/state/e2e-image-tag.txt`
+	export IMAGE=`cat devenv/state/e2e-image-tag.txt` && cd devenv && envsubst < e2e-tester.yaml > state/e2e-tester-formatted.yaml
+
+# deploy e2e job
+deploy-e2e: push-tester-image
+	cd devenv && /usr/bin/env sh scripts/deploy_e2e_tester.sh
+
+# to be run by e2e job for private cluster
+run-e2e:
```
OliverMKing, line 40: should this depend on the deploy-e2e and push-tester-image ones? I'd like to be able to run make e2e or make run-e2e to run an e2e test. Right now it looks like I'd have to run

make deploy-e2e then make run-e2e

Makefile
```diff
@@ -1,18 +1,41 @@
...
-e2e:
-	go test -v --count=1 --tags=e2e ./e2e
+dev-private-cluster:
+	terraform --version
+	cd devenv && mkdir -p state && cd private_cluster_tf && terraform init && terraform apply -auto-approve
+	cd devenv && /usr/bin/env sh scripts/deploy_addon_private_cluster.sh
 
-push:
+# aka make push (formerly known as make push-private)
+update-image-on-deployment:
```
OliverMKing, line 20: the way we actually use this is becoming a lot more complicated. comments or docs explaining how to use these would be good.

this is becoming a lot more complicated than just 
make dev
make push
make e2e

OliverMKing (changes requested):

Makefile
jaiveerk: makes sense!

Makefile
jaiveerk: yep good call! I'll add a README

Makefile
jaiveerk: you don't have to run make run-e2e. that gets run by the containerized tester. you would just run deploy-e2e. I'll document this in the README I mentioned in my reply to the other comment, I can definitely see that this is pretty confusing with no other context.

devenv/scripts/push_addon_image.sh
jaiveerk: update: after discussion, switching to this approach of merging public e2e changes to minimize size of this PR

README.md
```diff
@@ -16,6 +16,23 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+This project leverages Terraform and the local user's Azure credentials to run an extensive E2E suite.
+
+### Public Cluster/Public DNS Zone
+The process to run an E2E test for public clusters is as follows: 
+1. Run `make clean-public` or `make clean-all` to clear any preexisting Terraform state for the public cluster dev environment.
+2. Run `make dev-public-cluster` to deploy all Azure resources necessary to run a full suite, including a cluster with the add-on enabled. IMPORTANT: this does not start the add-on. The next step needs to be run for the add-on to be fully deployed with the correct image.
+3. Run `make update-image-on-deployment` to build the add-on image according to the user's local state/branch and push it to the add-on deployment. This step can be re-run when changes to the local add-on are made and the user wishes to manually test those changes on a cluster.
+4. Run `make deploy-e2e` to deploy the e2e tester job, which will run the e2e test suite inside the cluster.
+
+### Public Cluster/Public DNS Zone
```
OliverMKing, line 29: should this say private?

README.md
```diff
@@ -16,6 +16,23 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+The process to run an E2E test for public clusters is as follows:
```
OliverMKing, line 30: this too seems like it should say private

jaiveerk: it absolutely should - noticed the typo, but separating this into two PRs, one for updating the public e2e tests/e2e framework, and then a second one for all things private DNS. Will look at the diff once the other is merged in and take care of all changes that need to get done.

PR #31, opened 2023-02-13
Decouple from nginx

OliverMKing:
This pr decouples WAR from nginx. It makes the addon more extensible for planned future changes. There are no breaking (or user affecting) changes in this pr and this passes the e2e tests.

This decouples by
- Removes references to nginx consts in reconcilers so the reconcilers are more dynamic which will let us support multiple configurations in the future. Instead of consts, config structs are used which are easily extendable in the future.
- Splits external dns provisioning from nginx ingress controller provisioning
- Splits ingress class provisioning from controller provisioning
- Shifts responsibility for starting nginx related resources to a nginx runnable rather than just starting everything from the base operator manager. This couples nginx related resources to the nginx package rather than the root-level operator
- Generalizes other reconcilers so they rely on interfaces rather than singular nginx consts


OliverMKing:
I'll squash merge this when we merge it in 😆

bfoley13 (commented):
This isn't a minor change, can you elaborate on the ways you accomplished your goals in the PR description

OliverMKing:
> This isn't a minor change, can you elaborate on the ways you accomplished your goals in the PR description

Updated to include the specific ways

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -30,18 +30,19 @@ import (
...
 type PlaceholderPodController struct {
-	client client.Client
-	config *config.Config
+	client     client.Client
+	config     *config.Config
+	ingConfigs []*manifests.NginxIngressConfig
 }
 
-func NewPlaceholderPodController(manager ctrl.Manager, conf *config.Config) error {
+func NewPlaceholderPodController(manager ctrl.Manager, conf *config.Config, ingConfigs []*manifests.NginxIngressConfig) error {
```
sabbour, line 38: Should we pass in an array of a more generic type instead of NginxIngressConfig, like an interface? Or is the plan to add additional configs (for example for traffic controller) via an additional parameter?

pkg/controller/nginx/nginx.go
```diff
@@ -0,0 +1,109 @@
...
+	objs := []client.Object{}
+	for _, config := range n.ingConfigs {
+		objs = append(objs, manifests.NginxIngressControllerResources(n.conf, n.self, config)...)
+	}
+
+	return ingress.NewIngressControllerReconciler(n.manager, objs, n.name)
+}
+
+func (n *nginx) addConcurrencyWatchdog() error {
+	return ingress.NewNginxConcurrencyWatchdog(n.manager, n.conf, n.ingConfigs)
```
jveski, line 92: Here's what I'm seeing:

- the controller entrypoint (controller.go) constructs an `nginx` controller
- the nginx controller is implemented in large part by generic constructs in the `ingress` package
- in the future, other ingress controllers can use these same constructs

Is this accurate? If it is, we should use an interface type for the args receiving `n.ingConfigs` since `ingress`/`keyvault`/`osm` functions shouldn't be coupled to the nginx config. The interface can then expose getter functions that will be common across ingress implementations.

pkg/controller/keyvault/placeholder_pod.go
OliverMKing: This is something I was planning on doing. Was juggling lots of thoughts so it slipped my mind. Will swap to this real quick :)

pkg/controller/nginx/nginx.go
jveski: Related: the keyvault, service, and osm controllers aren't related to nginx directly (i.e. they'll work with any ingress implementation). So they should be constructed in controller.go. The problem is getting the ingress configs to them, maybe a getter on the nginx controller would be alright?

jveski (changes requested):

pkg/controller/keyvault/placeholder_pod.go
OliverMKing: Added interfaces in a few places. Don't think we need one for the IngressReconciler (this feature will be phased out and we should only generate one Ingress per service so it doesn't make sense to generalize). The watchdog is pretty specific to nginx right now so can't generalize that either.

pkg/manifests/external_dns.go
```diff
@@ -0,0 +1,213 @@
...
+			Kind:      "ServiceAccount",
+			Name:      dnsConfig.ResourceName,
+			Namespace: conf.NS,
+		}},
+	}
+}
+
+func newExternalDNSConfigMap(conf *config.Config, externalDnsConfig *ExternalDnsConfig) (*corev1.ConfigMap, string) {
+	js, err := json.Marshal(&map[string]interface{}{
+		"tenantId":                    externalDnsConfig.TenantId,
```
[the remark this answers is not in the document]
OliverMKing: Not opposed to that but this one was trying to compensate for future changes (we are planning on supporting multiple external dns instances in the future). In the case that we have that future we would in theory change the config object to support that as well but the problem is with the naming. 

We'd be taking something like 

privateExternalDnsConfigs []{tenantId, subId, etc}
publicExternalDnsConfigs []{tenantId, subId, etc}

in from the config but we'd need to assign a unique name to each one. Bundling that in a separate struct makes more sense to me than trying to parse that and figure out what the proper name should be based on the config in this function

pkg/controller/osm/ingress_backend_reconciler.go
```diff
@@ -71,7 +72,15 @@ func (i *IngressBackendReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		},
 	}
 
-	if ing.Annotations == nil || ing.Annotations["kubernetes.azure.com/use-osm-mtls"] == "" {
+	var controllerName string
+	for _, ingConfig := range i.ingConfigs {
+		if ing.Spec.IngressClassName != nil && ingConfig.IcName == *ing.Spec.IngressClassName {
+			controllerName = ingConfig.ResourceName
```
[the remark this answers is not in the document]
OliverMKing: Just updated that to do this + swapped to a generic

pkg/manifests/nginx_test.go
```diff
@@ -0,0 +1,155 @@
...
+func TestIngressClassResources(t *testing.T) {
+	for _, tc := range classTestCases {
+		objs := NginxIngressClass(tc.Conf, tc.Deploy, tc.IngConfig)
+		fixture := path.Join("fixtures", "nginx", tc.Name) + "-ingressclass.json"
+		AssertFixture(t, fixture, objs)
+	}
+}
+
+// AssertFixture checks the fixture path and compares it to the provided objects, failing if they are not equal
+func AssertFixture(t *testing.T, fixturePath string, objs []client.Object) {
```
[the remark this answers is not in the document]
OliverMKing: done

pkg/manifests/external_dns.go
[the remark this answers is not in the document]
jveski: Great point — let's keep as-is!

pkg/controller/nginx/nginx.go
OliverMKing: This is correct! I swapped to interfaces for the keyvault and osm functions. Will do the same for `NewNginxConcurrencyWatchdog` tomorrow (that's more work since it has some nginx specific things like metrics port and family.


A getter function on the nginx controller seems like the only way to get the ingress configs to them, unfortunately. I'd prefer that to passing them as an argument to the controller.

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -111,8 +114,8 @@ func (i *IngressSecretProviderClassReconciler) buildSPC(ing *netv1.Ingress, spc
 	}
 
 	managed := false
-	for _, ingConfig := range i.ingConfigs {
-		if ing.Spec.IngressClassName != nil && ingConfig.IcName == *ing.Spec.IngressClassName {
+	for _, manager := range i.managers {
+		if manager.IsManaging(ing) {
```
jveski, line 119: very nice 👍

pkg/manifests/nginx.go
```diff
@@ -32,6 +32,31 @@ func (n *NginxIngressConfig) PodLabels() map[string]string {
...
+func (n *NginxIngressConfig) IsManaging(ing *netv1.Ingress) bool {
+	if ing.Spec.IngressClassName == nil {
+		return false
+	}
+
+	if *ing.Spec.IngressClassName != n.IcName {
+		return false
+	}
+
+	return true
```
jveski, line 44: nit: `return ng.Spec.IngressClassName != nil && *ing.Spec.IngressClassName == n.IcName`

pkg/manifests/nginx.go
```diff
@@ -32,6 +32,31 @@ func (n *NginxIngressConfig) PodLabels() map[string]string {
...
+}
+
+// IngressControllerName returns the name of the controller deployment and a boolean indicating if the Ingress consumes that controller
+func (n *NginxIngressConfig) IngressControllerName(ing *netv1.Ingress) (string, bool) {
```
jveski, line 48: I'm wondering if it would be clearer to have a function like `DeploymentName()`? The caller would first call `IsManaging`, then if true call `DeploymentName`. A bit dry-er but more logic on the caller side so I'm not sold either way

OliverMKing: Made an updated to this to just internally call IsManaging first. I'm not sold either way on this either

pkg/controller/nginx/nginx.go
OliverMKing: Swapped all to generalized + moved the common ones back to the top-level controller

OliverMKing:
@jveski, @bfoley13, @sabbour I appreciate the feedback. I've addressed everything + made changes. I really like where this PR is at right now so I'd love one final round of reviews so we can get this merged in.

pkg/controller/controller.go
```diff
@@ -68,28 +68,48 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
...
+
+	ingressManagers := make([]keyvault.IngressManager, 0)
+	for _, nginxConfig := range nginxConfigs {
+		ingressManagers = append(ingressManagers, nginxConfig)
+	}
+	if err := keyvault.NewIngressSecretProviderClassReconciler(m, conf, ingressManagers); err != nil {
 		return nil, err
 	}
-	if err = keyvault.NewEventMirror(m, conf); err != nil {
+	if err := keyvault.NewPlaceholderPodController(m, conf, ingressManagers); err != nil {
```
jaiveerk, line 98: more of a question than feedback - why instantiate a new error object here but then overwrite it in the next statement?

pkg/controller/controller.go
```diff
@@ -68,28 +68,48 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
...
+	nginxConfigs, err := nginx.New(m, conf, deploy)
+	if err != nil {
 		return nil, err
 	}
-	if err = keyvault.NewIngressSecretProviderClassReconciler(m, conf); err != nil {
+
+	watchdogTargets := make([]*ingress.WatchdogTarget, 0)
+	for _, nginxConfig := range nginxConfigs {
+		watchdogTargets = append(watchdogTargets, &ingress.WatchdogTarget{
+			ScrapeFn:    ingress.NginxScrapeFn,
```
jaiveerk, line 83: Might be missing something in the larger scheme of things but are we ever going to be creating watchdog targets for nginx objects that DON'T use the nginx scrape function? Why import the scrape function separately instead of just including it in the config struct? That way we can ensure unique configs can all have their own scrape function associated with them rather than making future contributors make that logical leap?

pkg/controller/ingress/concurrency_watchdog.go
```diff
@@ -24,10 +25,57 @@ import (
...
 	"github.com/Azure/aks-app-routing-operator/pkg/config"
-	"github.com/Azure/aks-app-routing-operator/pkg/manifests"
 	"github.com/Azure/aks-app-routing-operator/pkg/util"
 )
 
+// ScrapeFn returns the connection count for the given pod
+type ScrapeFn func(ctx context.Context, client rest.Interface, pod *corev1.Pod) (float64, error)
+
+// NginxScrapeFn is the scrape function for Nginx
+func NginxScrapeFn(ctx context.Context, client rest.Interface, pod *corev1.Pod) (float64, error) {
```
jaiveerk, line 35: follow up on prev comment but why separate this from nginx config? this would only ever be used for an nginx ingress controller right?

pkg/controller/ingress/ingress_class_reconciler.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/common"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewIngressClassReconciler creates a reconciler that manages ingress class resources
+func NewIngressClassReconciler(manager ctrl.Manager, resources []client.Object, name string) error {
+	return common.NewResourceReconciler(manager, name+"IngressClassReconciler", resources, reconcileInterval)
```
jaiveerk, line 11: this is awesome, super clean

pkg/manifests/external_dns.go
```diff
@@ -0,0 +1,213 @@
...
+						"checksum/configmap": configMapHash[:16],
+					},
+				},
+				Spec: *WithPreferSystemNodes(&corev1.PodSpec{
+					ServiceAccountName: externalDnsConfig.ResourceName,
+					Containers: []corev1.Container{*withLivenessProbeMatchingReadiness(withTypicalReadinessProbe(7979, &corev1.Container{
+						Name:  "controller",
+						Image: path.Join(conf.Registry, "/oss/kubernetes/external-dns:v0.11.0.2"),
+						Args: []string{
+							"--provider=azure",
```
jaiveerk, line 177: wanted to change these into consts but didn't since you mentioned reorganizing these (which is awesome btw, this is SO much easier to read), but could you put stuff like azure and the "ingress" source as constants now that the placement of these strings is more stable?

jaiveerk (changes requested):

pkg/controller/controller.go
OliverMKing: What do you mean by import the scrape function separately? Each unique config does have their own scrape function.

Are you suggesting we should have an additional nginx config struct that automatically has that?

pkg/controller/controller.go
OliverMKing: it doesn't really matter. It's not overwritten because := is scoped to whatever block it's in. The err is exclusive to that if statement.

pkg/controller/ingress/concurrency_watchdog.go
OliverMKing: This was never connected to the nginx config. This is a different abstraction. This package has no concept of the nginx config, just LabelGetters.

pkg/manifests/external_dns.go
OliverMKing: what is the value of making these consts? I don't see the value of making "azure" a const because it is extremely unlikely that will ever change, provider is only referenced in this single place, and what it means is most clear leaving it where it's used.

Source possibly makes more sense to move to a const since it's reasonable that we'd add new ones in the future (gateway potentially). However, the usage for source is that you pass each one in with a separate `--source` flag (https://github.com/kubernetes-sigs/external-dns/blob/master/pkg/apis/externaldns/types.go#L406). That would mean your const would actually be an array and you'd have to add logic here handling that making it more difficult to read than the nicely formatted string we see below. I also don't see any advantages to doing this for the same reasons as the azure const so it just would hurt code readability with no real benefits.

pkg/controller/controller.go
OliverMKing: I do tend to prefer := because I think block scoping things is really good. It limits where you'd have to look and reason about in the case of a bug.

pkg/controller/controller.go
OliverMKing: Saw the later comment. I'm assuming you mean attaching it to the nginxIngressConfig struct that lives in the manifests. Keeping them separate simply allows for more flexibility in the future. Things are more flexible and less tightly coupled like this. It also gives an example of what a scrape fn would actually look like in the package it's used in which helps anyone reading that file understand.

I don't think it's a large logical leap. It's very logical that nginxConfigs would want to use the nginxScrapeFn.

jveski (dismissed):
Looks good at a high level. I don't have time for a detailed review, but I like the general approach so I'll approve and let y'all work out the details. :ship:

pkg/manifests/external_dns.go
jaiveerk: rather than for future-proofing for change, I think it's just easier in case that string ends up being used elsewhere, like in other resources we end up deploying. 

Just a thought, cool to just resolve this as well if you feel necessary

pkg/controller/controller.go
jaiveerk: Ok, that makes sense. I see what you mean about flexibility but I personally still think that coupling the scrape function would make the separation from nginx cleaner along with making it easier to consolidate all the requirements/required functionality for if/when we add a new ingress provider. But if you feel strongly about the design here then I trust your understanding.

pkg/controller/ingress/ingress_controller_reconciler.go
```diff
@@ -4,71 +4,16 @@
...
-	}
-}
-
-func (i *IngressControllerReconciler) tick(ctx context.Context) error {
-	start := time.Now()
-	i.logger.Info("starting to reconcile ingress controller resources")
-	defer func() {
-		i.logger.Info("finished reconciling ingress controller resources", "latencySec", time.Since(start).Seconds())
-	}()
+const reconcileInterval = time.Minute * 3
```
bfoley13, line 14: any reason for the increase?

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -0,0 +1,37 @@
...
+const reconcileInterval = time.Minute * 3
+
+// newExternalDnsReconciler creates a reconciler that manages external dns resources
+func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) error {
+	return common.NewResourceReconciler(manager, "externalDnsReconciler", resources, reconcileInterval)
+}
+
+// NewExternalDns starts all resources required for external dns
+func NewExternalDns(manager ctrl.Manager, conf *config.Config, self *appsv1.Deployment) error {
+	if conf.DNSZoneDomain == "" {
```
bfoley13, line 23: Is this currently the only thing we need to check?

OliverMKing: Yes. It matches how WAR currently works.

We could validate that tenantId, subscription, resource group, domain, etc are valid in the config validate function but currently RP is responsible for that before they pass that information to us.

Anything you think we should check?

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -89,7 +88,14 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
 			return ctrl.Result{}, err
 		}
 	}
-	if ing.Name == "" || ing.Spec.IngressClassName == nil || *ing.Spec.IngressClassName != manifests.IngressClass {
+
+	managed := false
+	for _, manager := range p.managers {
```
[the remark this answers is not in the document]
OliverMKing: We should be able to swap to using a set for O(1). Will do that where possible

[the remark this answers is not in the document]
OliverMKing: Will update managers and ingressControllerNamer

bfoley13 (commented):
Very clean! A lot of great changes in here that will make things so much simpler in the future!

pkg/controller/dns/external_dns_reconciler.go
bfoley13: If its settled in RP then we should be good!

bfoley13 (approved):

jaiveerk (dismissed):

bfoley13 (approved):
lgtm!

jaiveerk (approved):
this is awesome

PR #33, opened 2023-02-15
Refactoring Public DNS Zones E2E

jaiveerk:
PR to include changes/refactor for public dns zone e2e testing to be compatible with testing framework for private dns zone e2e.


Makefile
```diff
@@ -1,18 +1,34 @@
...
-	kubectl set image -n kube-system --kubeconfig devenv/state/kubeconfig deployments/app-routing-operator operator=`cat devenv/state/operator-image-tag.txt`
+	cd devenv && /usr/bin/env sh scripts/push_image.sh
+
+push-tester-image:
+	# grab the image name and tag - that first gets determined by where and how we push it - so first need to build and push it - then export that name and registry and tag into the YAML
+	az acr login -n `cat devenv/state/registry.txt`
+	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator-e2e:$(shell date +%s)" > devenv/state/e2e-image-tag.txt
+	docker build --platform=linux/amd64 -t `cat devenv/state/e2e-image-tag.txt` -f e2e/Dockerfile .
+	docker push `cat devenv/state/e2e-image-tag.txt`
+	export IMAGE=`cat devenv/state/e2e-image-tag.txt` && cd devenv && envsubst < e2e-tester.yaml > state/e2e-tester-formatted.yaml
```
OliverMKing, line 26: Using kustomize instead of this would make sense. That's what that tool is for and it's more flexible + native than bash

e2e/fixtures/client/main.go
```diff
@@ -21,8 +21,14 @@ func main() {
...
 		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
 			d := net.Dialer{Timeout: time.Second}
-			ns := nameservers[rand.Intn(len(nameservers)-1)]
-			ns = ns[:len(ns)-1] // remove trailing period added for some reason by azure dns
+			var ns string
+			if len(nameservers) > 1 {
+				ns = nameservers[rand.Intn(len(nameservers)-1)]
+				ns = ns[:len(ns)-1] // remove trailing period added for some reason by azure dns
+			} else {
+				ns = nameservers[0]
```
OliverMKing, line 29: why do we not remove the trailing period in the else as well? we should probably do it after the if else

jaiveerk: it has to do with the kind of value we're feeding in. if multiple nameservers are fed in, we know that's coming from a public DNS zone's NS server info, and the addresses for those contain a dot at the end.

If there's only one address fed in, we know it's referring to the NS server of a private cluster, and so there's no trailing dot we'd need to remove.

Makefile
jaiveerk: I feel like leaving it in bash is easier to read/understand what's happening without needing an external tool, plus I feel like kustomize is really useful when there's shared state across lots of different files but since we're only doing this for two files I'm not sure that the overhead of setting up kustomize for this is worth it.

Also, if I'm not mistaken, we'd just need to write the same values to kustomization files that would hold the values, and we'd need the same kind of bash for that too right?

I'm not super familiar with kustomize though, reading up on it now. Can look more into it if you feel strongly about this!

bfoley13: I'm okay with leaving this for now but I think Oliver makes a good point here, we can probably simplify this in the future!

e2e/fixtures/client/main.go
OliverMKing: documenting that with a comment would be good

Makefile
jaiveerk: makes sense - I totally see the appeal and have something partially working, will try spending a bit longer on it and see how far I can get

e2e/fixtures/client/main.go
jaiveerk: will do!

Makefile
OliverMKing: Yeah this is fine for now but something good to keep in mind for the future

e2e/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.19
+
+WORKDIR /go/src/github.com/Azure/aks-app-routing-operator
+ADD ../. .
```
jveski, line 4: If this runs after Terraform (my understanding is that it does) then the image will contain the test cluster's kubeconfig.

Can we pass any configuration necessary to run the tests through the test job via environment variables or a configmap mount? And use the in cluster config to reach apiserver instead of the kubeconfig.

jveski: Oh- it would also be great to build the test binary in a build container and push only it rather than also pushing the source code.

Makefile
```diff
@@ -1,18 +1,34 @@
...
 
-push:
+# aka make push (formerly known as make push-private)
+update-image-on-deployment:
 	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator:$(shell date +%s)" > devenv/state/operator-image-tag.txt
 	az acr login -n `cat devenv/state/registry.txt`
 	docker build -t `cat devenv/state/operator-image-tag.txt` .
 	docker push `cat devenv/state/operator-image-tag.txt`
-	kubectl set image -n kube-system --kubeconfig devenv/state/kubeconfig deployments/app-routing-operator operator=`cat devenv/state/operator-image-tag.txt`
+	cd devenv && /usr/bin/env sh scripts/push_image.sh
```
jveski, line 18: Consider adding a shebang to the script instead of explicitly invoking a particular shell here.

jveski: Same for other scripts.

devenv/scripts/deploy_e2e_tester.sh
```diff
@@ -0,0 +1,40 @@
+CLUSTER_RESOURCE_GROUP=$(cat state/cluster-info.json | jq '.ClusterResourceGroup' | tr -d '"')
+CLUSTER_NAME=$(cat state/cluster-info.json | jq '.ClusterName' | tr -d '"')
+
+echo "deleting any existing e2e job..."
+az aks command invoke --resource-group $CLUSTER_RESOURCE_GROUP --name $CLUSTER_NAME --command "kubectl delete jobs app-routing-operator-e2e -n kube-system"
```
jveski, line 5: If you add `--ignore-not-found` to this command you can exec it along with the next `kubectl apply` command to avoid a second run command roundtrip (it's kind of slow).

Makefile
jaiveerk: good call! will do.

e2e/Dockerfile
jaiveerk: for the first bit - we can just remove the kubeconfig entirely - the containerized tester already uses the in cluster config, see [here](https://github.com/Azure/aks-app-routing-operator/pull/33/files#diff-e91d29c6d4c58503d9c2d72e2b04f1b70901a613e23008d1d1cfe8beff7038d4R47).

As far as I can tell, the only config that the e2e config needs is the e2e.json file. How would you feel about a go-embed of that JSON as opposed to creating a configmap? That way we're not importing extra stuff into the binary that just gets built once, but saving the step of having to write that info into a configmap and then deploying that

jaiveerk: actually - looks like because of the file structure, embed wouldn't be the cleanest solution. working on using a configmap now.

e2e/fixtures/client/main.go
sabbour: FWIW, trailing periods at the end of nameservers is documented in [RFC1034](https://www.rfc-editor.org/rfc/rfc1034). It isn't a quirky Azure DNS behavior.

Makefile
jaiveerk: update: using kustomize now, so marking this as resolved

Makefile
```diff
@@ -1,18 +1,33 @@
...
 
-clean:
-	rm -rf devenv/state devenv/.terraform.lock.hcl devenv/.terraform devenv/terraform.tfstate devenv/terraform.tfstate.backup
+# keep separate for simultaneous public/private dev without need for resource recreation
+clean-public:
+	rm -rf devenv/state devenv/public_cluster_tf/.terraform.lock.hcl devenv/public_cluster_tf/.terraform devenv/public_cluster_tf/terraform.tfstate devenv/public_cluster_tf/terraform.tfstate.backup
 
-e2e:
-	go test -v --count=1 --tags=e2e ./e2e
+dev-public-cluster:
```
OliverMKing, line 8: this should just be `dev-public`. Would match `clean-public` and shorter is better because we are running these commands all the time.

Makefile
```diff
@@ -1,18 +1,33 @@
...
+	terraform --version
+	cd devenv && mkdir -p state && cd public_cluster_tf && terraform init && terraform apply -auto-approve
 
-push:
+# aka make push (formerly known as make push-private)
```
OliverMKing, line 12: this comment doesn't make sense in the new code

Makefile
```diff
@@ -1,18 +1,33 @@
...
+update-image-on-deployment:
 	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator:$(shell date +%s)" > devenv/state/operator-image-tag.txt
 	az acr login -n `cat devenv/state/registry.txt`
 	docker build -t `cat devenv/state/operator-image-tag.txt` .
 	docker push `cat devenv/state/operator-image-tag.txt`
-	kubectl set image -n kube-system --kubeconfig devenv/state/kubeconfig deployments/app-routing-operator operator=`cat devenv/state/operator-image-tag.txt`
+	cd devenv && ./scripts/push_image.sh
```
OliverMKing, line 18: should probably just rename to `update-image`. Short and concise. 

The script you're calling should be renamed as well. Right now it's confusing because you're calling the push image script inside update image even though there's also push-tester-image below.

Makefile
```diff
@@ -1,18 +1,33 @@
...
+
+push-tester-image:
```
OliverMKing, line 20: rename to `push-image`. Short and concise

README.md
```diff
@@ -16,6 +16,16 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or
 contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
 
+## E2E
+This project leverages Terraform and the local user's Azure credentials to run an extensive E2E suite.
+
+### Public Cluster/Public DNS Zone
+The process to run an E2E test for public clusters is as follows: 
+1. Run `make clean-public` or `make clean-all` to clear any preexisting Terraform state for the public cluster dev environment.
```
OliverMKing, line 24: `make clean-call` doesn't exist right now

devenv/public_cluster_tf/main.tf
```diff
@@ -75,6 +75,11 @@ data "azurerm_user_assigned_identity" "clusteridentity" {
   resource_group_name = azurerm_kubernetes_cluster.cluster.node_resource_group
 }
 
+data "azurerm_resources" "noderesourcegroup" {
```
OliverMKing, line 78: how is this used?

devenv/scripts/deploy_e2e_tester.sh
```diff
@@ -0,0 +1,45 @@
...
+CLUSTER_RESOURCE_GROUP=$(cat devenv/state/cluster-info.json | jq '.ClusterResourceGroup' | tr -d '"')
+CLUSTER_NAME=$(cat devenv/state/cluster-info.json | jq '.ClusterName' | tr -d '"')
+
+echo "adding image tag to kustomize and generating configmap..."
+cp devenv/kustomize/* devenv/state/kustomize
+cd devenv/state/kustomize
+kustomize edit set image placeholderfortesterimage=`cat ../e2e-image-tag.txt`
+
+echo "deleting any existing e2e job..."
+az aks command invoke --resource-group $CLUSTER_RESOURCE_GROUP --name $CLUSTER_NAME --command "kubectl delete jobs app-routing-operator-e2e -n kube-system --ignore-not-found && kubectl apply -k ." --file .
```
OliverMKing, line 12: good to just separate the delete and creation parts into two separate invokes. It would be more obvious what was failing if one did fail that way

e2e/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.19 as builder
+
+WORKDIR /go/src/github.com/Azure/aks-app-routing-operator
+ADD ../. .
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go test -c -o e2e/e2e.test --count=1 --tags=e2e -c ./e2e
+
+FROM scratch
```
OliverMKing, line 7: pretty surprised this works with scratch. pretty cool!

e2e/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.19 as builder
+
+WORKDIR /go/src/github.com/Azure/aks-app-routing-operator
+ADD ../. .
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go test -c -o e2e/e2e.test --count=1 --tags=e2e -c ./e2e
```
OliverMKing, line 5: not sure if you've tried but if we don't need CGO_ENABLED=0 we should remove it. It slows done build times a lot

e2e/e2e_test.go
```diff
@@ -26,27 +27,33 @@ var (
...
 }
 
 func TestMain(m *testing.M) {
 	// Load configuration
-	rawConf, err := ioutil.ReadFile("../devenv/state/e2e.json")
-	if err != nil {
-		panic(err)
+	rawConf := os.Getenv("E2E_JSON_CONTENTS")
+	if rawConf == "" {
+		panic(fmt.Errorf("failed to get e2e contents from env"))
```
OliverMKing, line 40: use errors.New instead of fmt.Errorf when you don't need to format any variables in

OliverMKing (changes requested):

Makefile
jaiveerk: makes sense

e2e/e2e_test.go
```diff
@@ -26,27 +27,33 @@ var (
...
-	rc, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-		&clientcmd.ClientConfigLoadingRules{ExplicitPath: conf.Kubeconfig},
-		&clientcmd.ConfigOverrides{}).ClientConfig()
+
+	// attempt to load in-cluster config first
+	rc, err := rest.InClusterConfig()
 	if err != nil {
-		panic(err)
+		rc, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: conf.Kubeconfig},
```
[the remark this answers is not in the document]
jaiveerk: that was code from the old way of instantiating the client, but I don't think it's needed anymore so just gonna get rid of it

e2e/e2e_test.go
```diff
@@ -26,27 +27,33 @@ var (
 )
 
 type testConfig struct {
-	TestNamservers            []string
+	TestNameservers           []string
 	Kubeconfig                string
```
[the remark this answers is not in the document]
jaiveerk: good call!

Makefile
jaiveerk: after discussion renaming to:
make push
make e2e

getting rid of make push tester image entirely!

devenv/public_cluster_tf/main.tf
jaiveerk: it's only used in private - looks like it was left over from the divergence. thanks for catching that!

devenv/scripts/deploy_e2e_tester.sh
jaiveerk: I'm not getting the status of the command invoke command. The status that I'm getting is what's output from the JSON describing the pod

OliverMKing: The statuses do matter because of `set -e`.

e2e/fixtures/fixtures.go
```diff
@@ -49,8 +53,14 @@ func NewClientDeployment(t *testing.T, host string, nameservers []string) *appsv
 }
 
 func NewGoDeployment(t testing.TB, name string) *appsv1.Deployment {
```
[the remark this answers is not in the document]
jaiveerk: done! thanks for making me reinforce my understanding of iota :D

devenv/scripts/deploy_e2e_tester.sh
jaiveerk: modifying accordingly after our discussion!

README.md
bfoley13: How was this resolved @jaiveerk ?

jaiveerk: removed it as part of my latest commit (just pushed)! I marked it as resolved when I made the change locally to make it easier to figure out which comments I was done addressing.

jaiveerk: [diff](https://github.com/Azure/aks-app-routing-operator/pull/33/commits/3e098e21b3668f66fb010afaa04b834e47f4226f#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

e2e/Dockerfile
bfoley13: Can I ask what part didn't work when removing them? Curious if its the multi-stage using `scratch`?

e2e/e2e_test.go
bfoley13: resolved but not updated?

bfoley13: ah was on an old version of the pr all good

README.md
bfoley13: yeah seems i was on an old version of the pr! thanks!

devenv/scripts/command_invoke_with_output.sh
```diff
@@ -0,0 +1,27 @@
...
+  KUBECTL_EXIT_CODE=$(echo $FORMATTED_RESULT | jq '.exitCode')
+
+  if [ $KUBECTL_EXIT_CODE != "0" ]
+  then
+    echo "Failed to run kubectl command $KUBECTL_CMD"
+    echo $KUBECTL_CMD_RESULT
+    exit $KUBECTL_EXIT_CODE
+  fi
+
+  echo $KUBECTL_CMD_RESULT
```
OliverMKing, line 25: this function is fantastic! A return is likely better than an echo here.

OliverMKing (commented):
One very small comment. Everything else looks great :)

devenv/scripts/command_invoke_with_output.sh
jaiveerk: cant return non-numeric values

OliverMKing: jaiveer showed me that bash doesn't support returning anything other than ints LOL

OliverMKing:
Unit tests failing. Looks like it's running e2e as a unit test. Will approve after that's fixed. Like the current state of this

OliverMKing (approved):
LGTM. Good stuff

jveski (approved):
Looks good 💯

devenv/kustomize/e2e-tester.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  template:
+    metadata:
+      labels:
+        app: app-routing-operator-e2e
+    spec:
+      containers:
+        - name: tester
+          image: placeholderfortesterimage
+          command: [ "/go/e2e.test" ]
```
OliverMKing, line 15: this should just do `./e2e.test` and the Dockerfile just builds to ./e2e.test instead of /go/e2e.test. Having it be in the go directory breaks go conventions. I'd also be okay making this go/bin/e2e.test

OliverMKing (approved):

devenv/kustomize/e2e-tester.yaml
jaiveerk: per our discussion resolving for future pr

bfoley13 (approved):
Looks good!

OliverMKing (approved):

e2e/Dockerfile
jaiveerk: resolving after our discussion

PR #40, opened 2023-02-24
fix self deploy bug

OliverMKing:
# Description

If war is unable to find its own deployment then currently it will return a malformed deployment object and nil for that function. We have checks later that check if the deployment object is nil and handle that case accordingly. Those rely on us returning nil if the deployment is not found.

nil, nil is often an anitpattern in go but it's not in this case because nil is the desired value if self isn't found. self being nil is a concept that's supported within all of the war code. The reason why nil, nil is an antipattern in go is because if you receive a nil error from a function that means you should be able to use the other values (and nil is usually not actually a desired state of other values). This is good because otherwise you'd need multiple nil checks versus just one against error. In this case nil is a valid state of self (it's possible for it not to exist) so this is a fine pattern.

## Type of change

# How Has This Been Tested?

Tested with unit tests.

# Checklist:


pkg/controller/controller.go
```diff
@@ -136,11 +139,12 @@ func checkNamespace(kcs kubernetes.Interface, conf *config.Config) error {
...
 }
 
-func getSelfDeploy(kcs kubernetes.Interface, conf *config.Config) (*appsv1.Deployment, error) {
+func getSelfDeploy(kcs kubernetes.Interface, conf *config.Config, log logr.Logger) (*appsv1.Deployment, error) {
 	deploy, err := kcs.AppsV1().Deployments(conf.NS).Get(context.Background(), conf.OperatorDeployment, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		// It's okay if we don't find the deployment - just skip setting ownership references
-		err = nil
+		// It's okay if we don't find the deployment - just skip setting ownership references latter
+		log.Info("self deploy not found")
```
Tatsinnit, line 146: 💡 error handling suggests- just some thoughts - by doing this we are initiating a code path where there is no error.

can we still return an error with the “self deploy not dogune message”

tagt way we don’t need special case handle things.

OliverMKing: we aren't changing the code path in terms of the error. since this originally set err = nil we never returned an error if the self deploy wasn't found.

we could return an error but that doesn't accurately represent this function in the context of web app routing. It's okay if the self deployment isn't found and the rest of web app routing handles that case with a nil check on the deploy object.

Wrapping the deployment struct with information on whether it was found would be the alternative solution but that's overkill when `nil` is considered "not existing" in go.

Tatsinnit: Sure, I think I will let you decide, either we should custom error or enum it for more easy understanding of this scenario. 

The ripple down effect for special cases in long run as interesting. Here we are logging a piece of information which feels like should be more upfront for user.

**but** again if don’t know the scenario so mentioned above because it kind of grocery form the way son code path are working. 

Usually using a nil return is fine but be careful of - Your code will repeat all these nil check every times you need to access pmethod of an instance, contaminating your code with many condition that have nothing to do with your feature.

bfoley13 (approved):

PR #46, opened 2023-03-14
add prometheus annotations to Ingress controllers

OliverMKing:
# Description

This allows users to easily consume web app routing prometheus metrics. It follows the guidance in the [nginx docs](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/monitoring.md).

It won't disrupt or change the experience for users not using Prometheus in any way.

## Type of change

We should document this feature on our next release with examples.

# How Has This Been Tested?

- [x] Tested locally
- [x] Unit tested
- [x] E2e tested

# Checklist:

### DIFF
--- a/pkg/manifests/nginx.go
+++ b/pkg/manifests/nginx.go
@@ -5,6 +5,7 @@ package manifests
 
 import (
 	"path"
+	"strconv"
 
 	appsv1 "k8s.io/api/apps/v1"
 	autov1 "k8s.io/api/autoscaling/v1"
@@ -23,6 +24,23 @@ import (
 
 const (
 	controllerImageTag = "v1.3.0"
+	prom               = "prometheus"
+)
+
+var (
+	promServicePort = corev1.ServicePort{
+		Name:       prom,
+		Port:       10254,
+		TargetPort: intstr.FromString(prom),
+	}
+	promPodPort = corev1.ContainerPort{
+		Name:          prom,
+		ContainerPort: promServicePort.Port,
+	}
+	promAnnotations = map[string]string{
+		"prometheus.io/scrape": "true",
+		"prometheus.io/port":   strconv.Itoa(int(promServicePort.Port)),
+	}
 )
 
 // NginxIngressConfig defines configuration options for required resources for an Ingress
@@ -201,6 +219,10 @@ func newNginxIngressControllerService(conf *config.Config, ingressConfig *NginxI
 		annotations["external-dns.alpha.kubernetes.io/internal-hostname"] = "clusterip." + hostname
 	}
 
+	for k, v := range promAnnotations {
+		annotations[k] = v
+	}
+
 	return &corev1.Service{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Service",
@@ -227,6 +249,7 @@ func newNginxIngressControllerService(conf *config.Config, ingressConfig *NginxI
 					Port:       443,
 					TargetPort: intstr.FromString("https"),
 				},
+				promServicePort,
 			},
 		},
 	}
@@ -237,6 +260,11 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 	if !conf.DisableOSM {
 		podAnnotations["openservicemesh.io/sidecar-injection"] = "enabled"
 	}
+
+	for k, v := range promAnnotations {
+		podAnnotations[k] = v
+	}
+
 	return &appsv1.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
@@ -282,6 +310,7 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 								Name:          "https",
 								ContainerPort: 8443,
 							},
+							promPodPort,
 						},
 						Resources: corev1.ResourceRequirements{
 							Requests: corev1.ResourceList{
--- a/pkg/manifests/fixtures/nginx/full.json
+++ b/pkg/manifests/fixtures/nginx/full.json
@@ -191,6 +191,10 @@
         "labels": {
           "app.kubernetes.io/managed-by": "aks-app-routing-operator"
         },
+        "annotations": {
+          "prometheus.io/port": "10254",
+          "prometheus.io/scrape": "true"
+        },
         "ownerReferences": [
           {
             "apiVersion": "apps/v1",
@@ -211,6 +215,11 @@
             "name": "https",
             "port": 443,
             "targetPort": "https"
+          },
+          {
+            "name": "prometheus",
+            "port": 10254,
+            "targetPort": "prometheus"
           }
         ],
         "selector": {
@@ -255,7 +264,9 @@
               "app": "nginx"
             },
             "annotations": {
-              "openservicemesh.io/sidecar-injection": "enabled"
+              "openservicemesh.io/sidecar-injection": "enabled",
+              "prometheus.io/port": "10254",
+              "prometheus.io/scrape": "true"
             }
           },
           "spec": {
@@ -281,6 +292,10 @@
                   {
                     "name": "https",
                     "containerPort": 8443
+                  },
+                  {
+                    "name": "prometheus",
+                    "containerPort": 10254
                   }
                 ],
                 "env": [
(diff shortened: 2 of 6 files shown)


e2e/e2e_test.go
```diff
@@ -101,3 +109,12 @@ func TestBasicServiceNoOSM(t *testing.T) {
...
 		svr, svc)
 }
+
+// TestPrometheus proves that users can consume Prometheus metrics emitted by our controllers
+func TestPrometheus(t *testing.T) {
+	t.Parallel()
+	tc := suite.StartTestCase(t)
+	ns := tc.NS()
+
+	tc.WithResources(append(fixtures.NewPrometheus(ns), fixtures.NewPrometheusClient(ns, conf.PromClientImage))...)
```
jaiveerk, line 119: does the prometheus instance not get deployed by default? shouldn't the add-on automatically take care of that?

e2e/fixtures/fixtures.go
```diff
@@ -91,7 +91,7 @@ func NewGoDeployment(t testing.TB, d DeploymentType) *appsv1.Deployment {
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{{
 						Name:  "container",
-						Image: "mcr.microsoft.com/oss/go/microsoft/golang:1.18",
```
jaiveerk, line 94: I tried using 1.20 for client/server in RP but the linter threw an error with it saying some bazel thing was incompatible beyond 1.19, so for the sake of having the same version as RP in our e2e tests maybe use 1.19 instead?

e2e/fixtures/promclient/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/Azure/aks-app-routing-operator/e2e/fixtures/promclient
+
+go 1.20
```
jaiveerk, line 3: see comment about 1.19

pkg/manifests/nginx.go
```diff
@@ -201,6 +219,10 @@ func newNginxIngressControllerService(conf *config.Config, ingressConfig *NginxI
 		annotations["external-dns.alpha.kubernetes.io/internal-hostname"] = "clusterip." + hostname
 	}
```
bfoley13, line 221: Just want to call out that user will not be able to edit this resource in our new design, which means it will always be applied. I'm okay with this as nothing really changes but want to check with @sabbour on if adding these annotations should be optional?

bfoley13 (commented):
Please add to the description of the PR how the new prom test integrate with existing tests and the failure scenarios covered. It would also be helpful, if possible, to see successful E2E runs from a local environment. Otherwise, nice PR!

jaiveerk (changes requested):

pkg/config/config.go
```diff
@@ -8,10 +8,15 @@ import (
 	"flag"
 )
 
+const (
+	// DefaultNs is the default namespace for the resources deployed by this operator
+	DefaultNs = "app-routing-system"
```
jaiveerk, line 13: much needed lol

pkg/manifests/nginx.go
OliverMKing: They can edit the resource still. Patch doesn't override all annotations (just the ones specified).

devenv/scripts/deploy_e2e_tester.sh
```diff
@@ -43,11 +44,11 @@ POD_LOGS_RESULT=$(run_invoke $CLUSTER_NAME $CLUSTER_RESOURCE_GROUP 'kubectl logs
...
 then
-  echo "TEST FAILED"
-else
   echo "TEST PASSED"
+  exit 0
 fi
 
-exit $FINALSTATUS
\ No newline at end of file
+echo "TEST FAILED"
```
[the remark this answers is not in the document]
OliverMKing: It's a better pattern. There was a bug with the script beforehand that made it pass if $FINALSTATUS wasn't set (equal to ""). The pass case should be the explicit case that's being tested for to avoid unforeseen things.

e2e/e2e_test.go
OliverMKing: Which add-on would deploy prometheus? Nginx-ingresses emit metrics which can be consumed by prometheus. To prove that with an e2e test we need to deploy a prometheus instance so we can consume the metrics emitted by Nginx-ingresses.

e2e/fixtures/fixtures.go
OliverMKing: We don't need to have the same Go version as RP

e2e/fixtures/promclient/main.go
```diff
@@ -0,0 +1,62 @@
...
+		targets, err := api.Targets(ctx)
+		if err != nil {
+			log.Printf("error listing prometheus targets: %s", err)
+			w.WriteHeader(500)
+			return
+		}
+
+		for _, target := range targets.Active {
+			if target.Labels[jobKey] == nginxJob {
+				log.Print("found active nginx-ingress target")
```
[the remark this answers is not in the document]
OliverMKing: I'm not sure what value that adds. Active targets means that it truly is connected to the nginx-ingress controller so metrics are guaranteed to be read from the target.

Makefile
```diff
@@ -26,15 +26,17 @@ push:
 	docker push `cat devenv/state/operator-image-tag.txt`
 	./devenv/scripts/push_image.sh
 
-push-tester-image:
-	# grab the image name and tag - that first gets determined by where and how we push it - so first need to build and push it - then export that name and registry and tag into the YAML
+push-tester-images:
 	az acr login -n `cat devenv/state/registry.txt`
+	echo "$(shell cat devenv/state/registry.txt)/e2e-prom-client:$(shell date +%s)" > devenv/state/e2e-prom-client.txt
 	echo "$(shell cat devenv/state/registry.txt)/app-routing-operator-e2e:$(shell date +%s)" > devenv/state/e2e-image-tag.txt
+	docker build -t `cat devenv/state/e2e-prom-client.txt` -f e2e/fixtures/promclient/Dockerfile ./e2e/fixtures/promclient/
```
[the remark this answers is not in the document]
OliverMKing: Done!

OliverMKing:
> Please add to the description of the PR how the new prom test integrate with existing tests and the failure scenarios covered. It would also be helpful, if possible, to see successful E2E runs from a local environment. Otherwise, nice PR!

Updated the description

e2e/fixtures/prometheus.go
```diff
@@ -0,0 +1,232 @@
...
+    target_label: __scheme__
+    regex: (https?)
+  - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
+    action: replace
+    target_label: __metrics_path__
+    regex: (.+)
+  - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
+    action: replace
+    target_label: __address__
+    regex: ([^:]+)(?::\d+)?;(\d+)
```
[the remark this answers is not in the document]
OliverMKing: I added a comment at the top. This is just a standard nginx-ingress regex. This line appends the port to the address but that isn't important versus the fact that this is just a standard prometheus config for ingress-nginx

devenv/scripts/deploy_e2e_tester.sh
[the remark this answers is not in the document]
jaiveerk: makes sense!

e2e/fixtures/promclient/main.go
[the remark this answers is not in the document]
OliverMKing: The current test checks that the connection is valid and metrics are flowing. It's not relying on nginx working as expected it's checking and ensuring that it is working as expected in the test.

e2e/fixtures/fixtures.go
OliverMKing: I don't think we should hurt the quality of testing in this repo because of a limitation in RP. Being on a later version is better here. RP will eventually have to update to match.

If our goal is to completely replicate the tests here inside RP we should import them from here instead of copy / pasting them. That's not our goal though. Our goal is to have e2e tests here that ensure the operator works as expected and in rp the e2e tests ensure that the entire process works as expected.

e2e/fixtures/promclient/main.go
```diff
@@ -0,0 +1,62 @@
...
+	"log"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+)
+
+// these need to be the same as the consts in ../prometheus.go
```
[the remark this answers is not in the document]
davidgamero: ah i see yep that makes sense

davidgamero (approved):
looks good!

PR #56, opened 2023-06-12
add e2e tests to pr

OliverMKing:
# Description

Makes e2e tests run on a pr. Essentially, there's two workflows for validating a pr. One that runs on a trusted branch and one that runs on a fork. Both update the same status so we are able to block PRs on that status.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

Tested on a forked repo with all scenarios

# Checklist:


OliverMKing:
Workflows on this run are going to be wonky because of how GitHub workflows work. This will work once merged in tho

.github/workflows/ok-to-test.yaml
```diff
@@ -0,0 +1,24 @@
+# File based on https://github.com/imjohnbo/ok-to-test/blob/master/.github/workflows/ok-to-test.yml
+name: Ok To Test
+
+on:
+  issue_comment:
+    types: [created]
```
bfoley13, line 6: Okay I think I might be forgetting but changes to this action wont be run until theyre merged into main correct? If thats the case then my other comment on the PAT doesnt matter

.github/workflows/ok-to-test.yaml
```diff
@@ -0,0 +1,24 @@
...
+    permissions:
+      pull-requests: write
+
+    # Only run for PRs, not issue comments
+    if: ${{ github.event.issue.pull_request }}
+    steps:
+      - name: slash command dispatch
+        uses: peter-evans/slash-command-dispatch@a28ee6cd74d5200f99e247ebc7b365c03ae0ef3c # v3.0.1
+        with:
+          token: ${{ secrets.GH_PAT }} # we aren't allowed to use secrets.GITHUB_TOKEN here
```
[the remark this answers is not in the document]
OliverMKing: https://github.com/peter-evans/slash-command-dispatch#token

It's because the secrets.GITHUB_TOKEN doesn't have the proper permissions needed. secrets.GITHUB_TOKEN is something automatically populated by the GitHub runner (not something we create) so it's preferred to use that when possible.

.github/workflows/ok-to-test.yaml
OliverMKing: Yeah nothing on this workflow will work yet until it's merged into main.

.github/workflows/pr-validate-fork.yaml
```diff
@@ -0,0 +1,105 @@
...
+              status: 'in_progress',
+              details_url: url,
+            });
+            return result;
+  e2e:
+    needs: setup
+    if:
+      github.event_name == 'repository_dispatch' &&
+      github.event.client_payload.slash_command.args.named.sha != '' &&
+      contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.args.named.sha)
```
bfoley13, line 47: How does this completely stop risk of race condition?

OliverMKing: Verifying this though. That's how `if` should work because it's the same mechanism for cancelling jobs.

bfoley13: I think that would probably be best, or if we can just make e2e run based off SHA if it is present

bfoley13: Also whats the biggest thing we worry about with malicious code, with respect to github secrets it would have to be changes to actions right?

OliverMKing: patched things up so it's airtight here. The biggest worry with malicious code with respect to github secrets are two things
1. they can extract secrets if not they can get us to execute malicious code. In the context of this repo this isn't really a worry. The secrets we store aren't truly "secrets" because we use oidc which is a secretless authentication method. Only our github repo can authenticate with Azure through oidc. An attacker could see these secrets (which are more of ids like subscription id, service principal id, etc) but not utilize them in any way
2. They can execute malicious code on our infrastructure. Our e2e provisions aks clusters and runs tests on them. An injection of malicious code would allow an attacker to run that code on our cluster. Even further, they could get the kubeconfig or provision other infrastructure themselves. That's really what we must avoid here and we do with the fixes I added.

.github/workflows/ok-to-test.yaml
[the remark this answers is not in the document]
OliverMKing: I take this back. The action documentation mimics what I said above but apparently using just the normal GH_TOKEN will work in our use case https://github.com/peter-evans/slash-command-dispatch/issues/147

bfoley13 (approved):
looks good, nice addition to have testing on the PR's now!

davidgamero (approved):
looks good to me, this is a great improvement

PR #57, opened 2023-06-12
Multiple DNS Zone Support

jaiveerk:
# Description
This PR is to add support for multiple DNS zones.

## Type of change

# How Has This Been Tested?
E2E and unit tests have been updated to test this functionality.

- [x] E2E suite and terraform remade to use multiple zones
- [x] Fixture tests updated 
- [x] Unit tests to test new CLI args

# Checklist:


devenv/scripts/run_private_cluster.sh
```diff
@@ -1,10 +1,6 @@
-set -e
-echo "beginning private cluster tests..."
-make dev CLUSTER_TYPE=private DNS_ZONE_TYPE=public
-make push
-make e2e
-
-make dev CLUSTER_TYPE=private DNS_ZONE_TYPE=private
-make push
-make e2e
+	set -e
```
OliverMKing, line 1: weird indentation

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -20,19 +25,71 @@ func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) e
...
-		ResourceName:  "external-dns",
-		TenantId:      conf.TenantID,
-		Subscription:  conf.DNSZoneSub,
-		ResourceGroup: conf.DNSZoneRG,
-		Domain:        conf.DNSZoneDomain,
-		RecordId:      conf.DNSRecordID,
-		IsPrivate:     conf.DNSZonePrivate,
-	}
-	objs := manifests.ExternalDnsResources(conf, self, dnsConfig)
+	configs := parseZoneIds(conf)
```
OliverMKing, line 32: parse the zone ids higher up. We should do this in the config object.

We will use the zones in other places so it makes sense to move the parsing up

pkg/controller/dns/external_dns_reconciler_test.go
```diff
@@ -0,0 +1,11 @@
+package dns
+
+var (
+	privateZoneOne = "/subscriptions/test-subscription/resourceGroups/test-rg-private/providers/Microsoft.Network/privatednszones/test-one.com"
```
OliverMKing, line 4: there's no tests actually here?

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -20,19 +25,71 @@ func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) e
...
+
+func parseZoneIds(conf *config.Config) (configs []*manifests.ExternalDnsConfig) {
+	var ret []*manifests.ExternalDnsConfig
+	var privateZones, publicZones []string
+	var privateSubscription, privateZoneRg, publicSubscription, publicZoneRg string
+
+	for _, zoneId := range conf.DNSZoneIDs {
+		parsedZone, err := azure.ParseResourceID(zoneId)
+		// this should be impossible
+		if err != nil {
```
OliverMKing, line 49: we should at least log a warning

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -20,19 +25,71 @@ func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) e
...
+			continue
+		}
+
+		if strings.EqualFold(parsedZone.ResourceType, config.PrivateZoneType) {
+			// it's a private zone
+			privateZones = append(privateZones, zoneId)
+			privateSubscription = parsedZone.SubscriptionID
```
OliverMKing, line 56: we don't guarantee that the subs and rgs are the same here for each type. We need to parse this information higher up in config.go to validate. In config.go we should separate by public and private as well.

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -20,19 +25,71 @@ func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) e
...
+			publicSubscription = parsedZone.SubscriptionID
+			publicZoneRg = parsedZone.ResourceGroup
+		}
+	}
+
+	if len(privateZones) > 0 {
+		ret = append(ret, generateConfig(conf, privateZones, privateSubscription, privateZoneRg, manifests.PrivateProvider))
+	}
+
+	if len(publicZones) > 0 {
```
OliverMKing, line 70: what's the plan for cleanup going to be? We need to cleanup old external dns instances if they go from having private zones to no private zones

pkg/manifests/external_dns.go
```diff
@@ -4,51 +4,68 @@ import (
...
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"path"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	provider        = "azure"
-	privateProvider = "azure-private-dns"
+	Provider        = "azure"
```
OliverMKing, line 22: if you are going to use these like an enum they should be a separate type.


```Go
type Provider int

const (
    Public Provider = iota
    Private
)

func (p Provider) String() string {
   switch p {
   case Public:
      return "azure"
   case Private:
      return "azure-private-dns"
   default:
       return ""
}
```

OliverMKing: This is how you should represent something like this in Go.

devenv/scripts/run_private_cluster.sh
jaiveerk: idk why that keeps happening lol

pkg/controller/dns/external_dns_reconciler.go
jaiveerk: do you mean load info about which zones are private/public and then the resource groups and subscriptions directly into the config object?

pkg/controller/dns/external_dns_reconciler_test.go
jaiveerk: this was meant to include tests for DNS Zone config generation, but moving most of that logic to config anyway so will put testing there - thanks for catching that!

jaiveerk: added testing and moved logic in next commit, marking as resolved

pkg/config/config.go
```diff
@@ -71,22 +73,36 @@ func (c *Config) Validate() error {
...
+
+		parsedID, err := azure.ParseResourceID(id)
+		if err != nil {
+			return fmt.Errorf("failed to parse DNS Zone ID %s: %s", id, err)
+		}
+
+		isPublicDnsZone := strings.EqualFold(parsedID.ResourceType, PublicZoneType)
+		isPrivateDnsZone := strings.EqualFold(parsedID.ResourceType, PrivateZoneType)
+
+		if !strings.EqualFold(parsedID.Provider, "Microsoft.Network") || (!isPublicDnsZone && !isPrivateDnsZone) {
```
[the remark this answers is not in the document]
jaiveerk: I don't think this makes sense - 
that would result in a lot of extra information being stored, namely the resourcegroup and subscription being stored repeatedly for every struct.
Instead, I think my new approach makes sense - we're storing the parsed information for easy access, but also storing IDs for wherever they're needed

pkg/controller/dns/external_dns_reconciler.go
jaiveerk: after discussion offline, returning error here

pkg/config/config.go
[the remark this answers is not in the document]
OliverMKing: You don't need to store the rg and subscription for every struct.

Your struct can look like

struct privatedns {
ids name[]
rg string
sub string
}

pkg/config/config.go
```diff
@@ -44,8 +47,9 @@ type Config struct {
 	DisableKeyvault                                   bool
 	MSIClientID, TenantID                             string
 	Cloud, Location                                   string
-	DNSZoneRG, DNSZoneSub, DNSZoneDomain, DNSRecordID string
-	DNSZonePrivate                                    bool
+	PrivateZoneIds, PublicZoneIds                     []string
```
OliverMKing, line 50: It doesn't make sense to have these be separate structures. Group similar information with a struct

struct PrivateZone {
ids []string
sub string
rg string
}

pkg/config/config.go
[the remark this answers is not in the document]
jaiveerk: ok yeah that makes sense I can move to that

pkg/manifests/external_dns.go
jaiveerk: done!

pkg/config/config.go
jaiveerk: done

pkg/controller/dns/external_dns_reconciler.go
jaiveerk: marking as resolved per our discussion offline to keep this for a separate PR

pkg/manifests/external_dns.go
```diff
@@ -184,13 +204,12 @@ func newExternalDNSDeployment(conf *config.Config, configMapHash string, externa
...
-						Args: []string{
-							"--provider=" + provider,
+						Args: append([]string{
+							"--provider=" + externalDnsConfig.Provider,
 							"--source=ingress",
 							"--interval=3m0s",
-							"--txt-owner-id=" + externalDnsConfig.RecordId,
-							"--domain-filter=" + externalDnsConfig.Domain,
-						},
+							"--txt-owner-id=" + operatorName,
```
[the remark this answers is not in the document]
jaiveerk: using the kube-system uid as a unique identifier per [this](https://github.com/kubernetes/kubernetes/issues/77487#issuecomment-489786023) comment I saw in a thread about this issue.  

My one concern is needing a fallback issue in case getting the ns fails but not really sure what to do about that

pkg/controller/dns/cleanup_runner.go
```diff
@@ -0,0 +1,29 @@
...
+func newCleanupRunner(manager ctrl.Manager, namesToDelete []string) error {
+	runner := &CleanupRunner{
+		client:        manager.GetClient(),
+		namesToDelete: namesToDelete,
+	}
+	return manager.Add(runner)
+}
+
+func (r *CleanupRunner) Start(ctx context.Context) error {
+	// deletion would take place here using manifests.K8sNameKey
```
bfoley13, line 26: I assume this is for external dns cleanup, are we holding off on that?

jaiveerk: yep! we will likely include that in a future PR as we're still fleshing out some of the details/implementation, but didn't want to just remove what had already been done since it doesn't interfere with current functionality

pkg/manifests/external_dns_test.go
```diff
@@ -7,66 +7,79 @@ import (
...
 		},
 	}
 )
 
 func TestExternalDnsResources(t *testing.T) {
 	for _, tc := range testCases {
-		objs := ExternalDnsResources(tc.Conf, tc.Deploy, tc.DnsConfig)
+		objs := []client.Object{}
+
+		objs = append(objs, ExternalDnsResources(tc.Conf, tc.Deploy, tc.DnsConfigs)...)
```
bfoley13, line 81: Change isnt necessary

pkg/manifests/common_test.go
```diff
@@ -19,11 +19,11 @@ var (
 		Name   string
 		Config *config.Config
 	}{
-		{Name: "namespace", Config: &config.Config{
+		{Name: "Namespace", Config: &config.Config{
 			NS: "test-namespace",
 		}},
 		{
-			Name: "another-namespace",
+			Name: "another-Namespace",
```
bfoley13, line 26: You replace all this?

pkg/manifests/external_dns.go
```diff
@@ -120,11 +168,11 @@ func newExternalDNSClusterRoleBinding(conf *config.Config, dnsConfig *ExternalDn
...
 
-func newExternalDNSConfigMap(conf *config.Config, externalDnsConfig *ExternalDnsConfig) (*corev1.ConfigMap, string) {
+func newExternalDNSConfigMap(conf *config.Config, dnsConfig *ExternalDnsConfig) (*corev1.ConfigMap, string) {
 	js, err := json.Marshal(&map[string]interface{}{
-		"tenantId":                    externalDnsConfig.TenantId,
-		"subscriptionId":              externalDnsConfig.Subscription,
-		"resourceGroup":               externalDnsConfig.ResourceGroup,
+		"tenantId":                    dnsConfig.TenantId,
+		"subscriptionId":              dnsConfig.Subscription,
+		"resourceGroup":               dnsConfig.ResourceGroup,
```
bfoley13, line 175: any reason for these changes? its also externalDnsConfig within other func params

pkg/manifests/external_dns_test.go
jaiveerk: thanks for catching that!

pkg/manifests/common_test.go
jaiveerk: not sure how/why I made this but getting rid of this

pkg/manifests/external_dns.go
jaiveerk: thanks for catching that! this went through several iterations and I guess ended up just being the same with different var names.

Changing to standardize the arg name to just be externalDnsConfig

pkg/config/config.go
```diff
@@ -71,22 +83,68 @@ func (c *Config) Validate() error {
...
+		}
+
+		if !strings.EqualFold(parsedZone.Provider, "Microsoft.Network") {
+			return fmt.Errorf("invalid resource provider %s from zone %s: resource ID must be a public or private DNS Zone resource ID from provider Microsoft.Network", parsedZone.Provider, zoneId)
+		}
+
+		if strings.EqualFold(parsedZone.ResourceType, PrivateZoneType) {
+			// it's a private zone
+			c.PrivateZoneConfig.ZoneIds = append(c.PrivateZoneConfig.ZoneIds, zoneId)
+			c.PrivateZoneConfig.Subscription = parsedZone.SubscriptionID
```
bfoley13, line 131: Looks like this overwrites the sub/rg, what restrictions should we validate here?

jaiveerk: adding validation to ensure all private zones are in the same rg and that all public zones are in the same rg

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+
+func (c *Config) ParseZoneIDs(zonesString string) error {
+
+	c.PrivateZoneConfig = &DnsZoneConfig{}
+	c.PublicZoneConfig = &DnsZoneConfig{}
+
+	DNSZoneIDs := strings.Split(zonesString, ",")
+	for _, zoneId := range DNSZoneIDs {
+		if zoneId == "" {
+			return errors.New("--dns-zone-ids must not contain empty strings")
```
OliverMKing, line 116: Just continue if it's "". Not worth failing off that because the intent is clear.

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+		}
+
+		parsedZone, err := azure.ParseResourceID(zoneId)
+		if err != nil {
+			return fmt.Errorf("error while parsing dns zone resource ID %s: %s", zoneId, err)
```
OliverMKing, line 121: Never include "error" in your Go error messages. Every logging suite includes that when printing an error. Your error will show as 

Error: error while parsing

It's much better to just have

Error: while parsing ...

Not to mention the benefits when chaining and wrapping errors

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+			// it's a public zone
+			if err := validateZoneId(parsedZone, c.PublicZoneConfig.Subscription, c.PrivateZoneConfig.ResourceGroup); err != nil {
+				return err
+			}
+
+			c.PublicZoneConfig.Subscription = parsedZone.SubscriptionID
+			c.PublicZoneConfig.ResourceGroup = parsedZone.ResourceGroup
+			c.PublicZoneConfig.ZoneIds = append(c.PublicZoneConfig.ZoneIds, zoneId)
+		} else {
+			return fmt.Errorf("error while parsing dns zone resource ID %s: detected invalid resource type %s", zoneId, parsedZone.ResourceType)
```
OliverMKing, line 148: Remove error

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+			// it's a private zone
+			if err := validateZoneId(parsedZone, c.PrivateZoneConfig.Subscription, c.PrivateZoneConfig.ResourceGroup); err != nil {
+				return err
+			}
+
+			c.PrivateZoneConfig.Subscription = parsedZone.SubscriptionID
+			c.PrivateZoneConfig.ResourceGroup = parsedZone.ResourceGroup
+			c.PrivateZoneConfig.ZoneIds = append(c.PrivateZoneConfig.ZoneIds, zoneId)
+
+		} else if strings.EqualFold(parsedZone.ResourceType, PublicZoneType) {
```
OliverMKing, line 138: Switch statement is more effective go than if elif else here

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+			return fmt.Errorf("error while parsing dns zone resource ID %s: detected invalid resource type %s", zoneId, parsedZone.ResourceType)
+		}
+	}
+
+	return nil
+}
+
+func validateZoneId(parsedZone azure.Resource, subscription, resourceGroup string) error {
+	if subscription != "" && parsedZone.SubscriptionID != subscription {
+		return fmt.Errorf("error while parsing resource IDs for %s: detected multiple subscriptions %s and %s", parsedZone.ResourceType, parsedZone.SubscriptionID, subscription)
```
OliverMKing, line 157: Remove "error" from return

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+	}
+
+	if resourceGroup != "" && parsedZone.ResourceGroup != resourceGroup {
+		return fmt.Errorf("error while parsing resource IDs for %s: detected multiple resource groups %s and %s", parsedZone.ResourceType, parsedZone.ResourceGroup, resourceGroup)
```
OliverMKing, line 161: Remove "error"

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
 		return errors.New("--concurrency-watchdog-threshold must be greater than 100")
 	}
 	if c.ConcurrencyWatchdogVotes < 1 {
 		return errors.New("--concurrency-watchdog-votes must be a positive number")
 	}
+
+	if c.ClusterFqdn == "" {
+		return errors.New("--cluster-fqdn is required")
+	} else {
+		c.ClusterFqdn = hash(c.ClusterFqdn)
```
OliverMKing, line 96: Why are we hashing if we are using the cluster fqdn?

pkg/controller/dns/cleanup_runner.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+type CleanupRunner struct {
+	client        client.Client
+	namesToDelete []string
```
OliverMKing, line 12: unusedNameLabels would be a better name for this field

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -20,19 +23,51 @@ func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) e
...
+	objs := append([]client.Object{}, manifests.ExternalDnsResources(conf, self, configs)...)
+
 	return newExternalDnsReconciler(manager, objs)
 }
+
+func generateZoneConfigs(conf *config.Config) (configs []*manifests.ExternalDnsConfig, namesToDelete []string) {
+	publicResourceName := fmt.Sprintf("%s%s", manifests.ExternalDnsResourceName, manifests.PublicSuffix)
+	privateResourceName := fmt.Sprintf("%s%s", manifests.ExternalDnsResourceName, manifests.PrivateSuffix)
+
+	if conf.PrivateZoneConfig != nil && len(conf.PrivateZoneConfig.ZoneIds) > 0 {
```
OliverMKing, line 48: All these nil checks make the code a little hard to read. Just make the config take the struct instead of a pointer to avoid this. There's basically no performance benefit to using a pointer to the zone config fields

pkg/controller/dns/external_dns_reconciler.go
```diff
@@ -20,19 +23,51 @@ func newExternalDnsReconciler(manager ctrl.Manager, resources []client.Object) e
...
+		configs = append(configs, generateConfig(conf, conf.PublicZoneConfig, manifests.PublicProvider, publicResourceName))
+	} else {
+		namesToDelete = append(namesToDelete, publicResourceName)
+	}
+
+	// namesToDelete will eventually be used for externalDNS cleanup
+	return
+}
+
+func generateConfig(conf *config.Config, dnsZoneConfig *config.DnsZoneConfig, provider manifests.Provider, resourceName string) *manifests.ExternalDnsConfig {
```
OliverMKing, line 64: Where is this used? On my phone so maybe I'm blind but I don't see this used anywhere

pkg/manifests/common.go
```diff
@@ -24,7 +26,7 @@ func getOwnerRefs(deploy *appsv1.Deployment) []metav1.OwnerReference {
 	}}
 }
 
-func namespace(conf *config.Config) *corev1.Namespace {
+func Namespace(conf *config.Config) *corev1.Namespace {
```
OliverMKing, line 29: Not sure why Namespace is being exported now?

bfoley13 (dismissed):
changes look good to me

pkg/config/config.go
```diff
@@ -35,21 +40,28 @@ func init() {
...
+	NS, Registry                        string
+	DisableKeyvault                     bool
+	MSIClientID, TenantID               string
+	Cloud, Location                     string
+	PrivateZoneConfig, PublicZoneConfig *DnsZoneConfig
+	ConcurrencyWatchdogThres            float64
+	ConcurrencyWatchdogVotes            int
+	DisableOSM                          bool
+	OperatorDeployment                  string
+	ClusterFqdn                         string
```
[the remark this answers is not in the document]
jaiveerk: ok, I'll do that and use the hash function only when creating the externalDnsConfig

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
 	}
+
+	if c.ClusterFqdn == "" {
+		return errors.New("--cluster-fqdn is required")
+	} else {
+		c.ClusterFqdn = hash(c.ClusterFqdn)
+	}
+
+	if dnsZonesString != "" {
+		if err := c.ParseZoneIDs(dnsZonesString); err != nil {
```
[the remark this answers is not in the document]
jaiveerk: what happens if parsing fails? parsing is, in a sense, validation, right?

pkg/config/config.go
```diff
@@ -35,21 +40,28 @@ func init() {
 	flag.StringVar(&Flags.MetricsAddr, "metrics-addr", "0.0.0.0:8081", "address to serve Prometheus metrics on")
 	flag.StringVar(&Flags.ProbeAddr, "probe-addr", "0.0.0.0:8080", "address to serve readiness/liveness probes on")
 	flag.StringVar(&Flags.OperatorDeployment, "operator-deployment", "app-routing-operator", "name of the operator's k8s deployment")
+	flag.StringVar(&Flags.ClusterFqdn, "cluster-fqdn", "", "fqdn of the cluster the add-on belongs to")
+}
+
+type DnsZoneConfig struct {
```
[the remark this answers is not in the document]
jaiveerk: it's used in externalDNS reconciler to generate externalDNS configs from the struct containing the information it needs.

previously had a design where fields were just directly in the config object but you suggested moving it to a struct within the config body

pkg/config/config.go
```diff
@@ -71,22 +83,89 @@ func (c *Config) Validate() error {
...
+
 	return nil
 }
+
+func (c *Config) ParseZoneIDs(zonesString string) error {
+
+	c.PrivateZoneConfig = &DnsZoneConfig{}
+	c.PublicZoneConfig = &DnsZoneConfig{}
+
+	DNSZoneIDs := strings.Split(zonesString, ",")
```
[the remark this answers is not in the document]
jaiveerk: these fields are used in external dns reconciler

pkg/config/config.go
jaiveerk: removed this and the test case associated with it!

pkg/config/config.go
jaiveerk: good call! getting rid of it here and anywhere else it's used

pkg/config/config.go
jaiveerk: done!

pkg/config/config.go
jaiveerk: figured it was just a more effective/concise way to store the data. do you feel strongly about using the fqdn specifically?

pkg/config/config.go
jaiveerk: switched! (pun fully intended)

pkg/controller/dns/cleanup_runner.go
jaiveerk: done

pkg/controller/dns/external_dns_reconciler.go
jaiveerk: done - goo call

pkg/manifests/common.go
jaiveerk: this was because of a change I had made in a different iteration but that's no longer needed - re converting!

pkg/controller/dns/external_dns_reconciler.go
jaiveerk: used to generate externalDNS configs from the conf object. line 30 of this file and it's tested in the external_dns_reconciler test file.

pkg/config/config.go
OliverMKing: Owner id is typically human readable

jaiveerk: sounds good ill remove the hashing

bfoley13 (dismissed):

OliverMKing (dismissed):

bosesuneha (approved):
Lgtm

PR #61, opened 2023-07-05
add cleanup e2e tests and enable cleanup

OliverMKing:
# Description

Adds e2e checks that verify cleanup functionality. Also enables the cleanup functionality.

## Type of change

# How Has This Been Tested?

Tested locally and e2e 

# Checklist:


e2e/e2e_test.go
```diff
@@ -57,10 +60,23 @@ type testConfig struct {
...
 }
 
-func TestMain(m *testing.M) {
+func (testConfig *testConfig) Validate() error {
+	if len(testConfig.PrivateDNSZoneIDs) == 0 {
+		return errors.New("missing private dns zone ids")
+	}
+
+	if len(testConfig.PublicDNSZoneIDs) == 0 {
+		return errors.New("missing public dns zone ids")
```
jaiveerk, line 73: can we modify this check to only fail if zero zones are provided? that way we can still test scenarios for only public and only private.

moreover - I'm not sure if we have tests that test our functionality when no zones are provided, and this would prevent us from doing that, so maybe we should get rid of this check completely? or perhaps add a warning?

jaiveerk (changes requested):
small bits of feedback but looks awesome!

e2e/e2e_test.go
OliverMKing: We shouldn't.

The way our e2e tests should work is by testing all scenario directly in e2e tests. The way you do that is by passing in all options into the config then changing the operator config in the tests when you want to check different configurations. That's how we test cleanup.

If we pass all options into the config then let the e2e testing handle all testing e2e is a lot more valuable and trustworthy. We should favor Go-based testing matrixes rather than bash which this accomplishes. Ideally all tests are bundled under one e2e so really the right thing to do is eventually take in both a private and public cluster in the e2e test config and kick off tests for them directly in Go.



On the zero zone tests and one zone scenarios those are tested in cleanup.

pkg/manifests/fixtures/external_dns/full.json
```diff
@@ -30,7 +29,7 @@
         "creationTimestamp": null,
         "labels": {
           "app.kubernetes.io/managed-by": "aks-app-routing-operator",
-          "app.kubernetes.io/name": "external-dns-private"
```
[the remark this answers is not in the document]
OliverMKing: All were being given the private suffix due to a weird Golang reference thing. Not particularly worried about it because it will all be fixed in reconciliation if something rolls out.

e2e/e2e_test.go
jaiveerk: makes sense! I agree with the Go matrix approach.

As for the zero and one cases - I may be missing something but it looks like there's no testing beyond just validating the externalDNS resources' existence right now. may be a good idea for a future task to add something of significance there to ensure the add-on still works as intended.

pkg/manifests/fixtures/external_dns/full.json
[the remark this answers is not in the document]
jaiveerk: makes sense

jaiveerk (approved):

e2e/e2e_test.go
OliverMKing: For sure something we can add in the future! The current tests do ensure that the operator is running correctly for 0 zone because if the operator fails to run cleanup won't occur for zero zone.

bfoley13 (approved):

PR #72, opened 2023-07-19
Adding File Name and Line Number to Logs

jaiveerk:
# Description

This PR is to modify the way we construct our zap logger to include caller information in each entry

## Type of change

# How Has This Been Tested?

- [x] Unit tests have been added to ensure that the caller is present
- [ ] Test B

# Checklist:


pkg/controller/controller.go
```diff
@@ -43,8 +45,14 @@ func init() {
...
 
 func getLogger(opts ...zap.Opts) logr.Logger {
+	// make raw logger object from options in original zap pkg
+	rawOpts := zap.RawZapOpts(ubzap.AddCaller())
+
+	// append k8s zap options to raw options
+	rawLogger := zap.NewRaw(append(opts, rawOpts)...)
+
 	// zap is the default recommended logger for controller-runtime when wanting json structured output
-	return zap.New(opts...)
```
OliverMKing, line 47: we would like to use the controller runtime zap logger. It has some k8s helpers

jaiveerk: this is the same thing done by the controller runtime zap logger:  https://github.com/kubernetes-sigs/controller-runtime/blob/116a1b831fffe7ccc3c8145306c3e1a3b1b14ffa/pkg/log/zap/zap.go#L42

pkg/controller/controller.go
```diff
@@ -43,8 +45,14 @@ func init() {
 }
 
 func getLogger(opts ...zap.Opts) logr.Logger {
+	// make raw logger object from options in original zap pkg
+	rawOpts := zap.RawZapOpts(ubzap.AddCaller())
+
+	// append k8s zap options to raw options
+	rawLogger := zap.NewRaw(append(opts, rawOpts)...)
```
OliverMKing, line 52: why use this NewRaw instead of the NewRaw in controller runtime?

pkg/controller/controller.go
```diff
@@ -43,8 +45,14 @@ func init() {
 }
 
 func getLogger(opts ...zap.Opts) logr.Logger {
+	// make raw logger object from options in original zap pkg
+	rawOpts := zap.RawZapOpts(ubzap.AddCaller())
```
OliverMKing, line 49: why can we not pass this in to the controlleruntime zap.New?

pkg/controller/controller.go
OliverMKing: yes but why are we using zapr.NewLogger now?

pkg/controller/controller.go
jaiveerk: no such option exists in the package. tried using a new [json encoderconfig](https://github.com/Azure/aks-app-routing-operator/pull/72/files#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1R28) with the caller key field populated but that didn't work

pkg/controller/controller.go
jaiveerk: it's how the controller runtime package converts the raw logger to the logr interface so I figured that was the best way to make that conversion

pkg/controller/controller.go
OliverMKing: image

pkg/controller/controller.go
OliverMKing: The parameters to NewRaw and zap.New are the same. We don't need a raw logger

pkg/controller/controller.go
OliverMKing: No need for rawLogger or zapr

pkg/controller/controller.go
jaiveerk: ah, thanks for catching that! I just saw the package was doing it and figured it was necessary but I see what you mean. fixing now

pkg/controller/controller.go
```diff
@@ -43,8 +44,11 @@ func init() {
 }
 
 func getLogger(opts ...zap.Opts) logr.Logger {
+	// make raw logger object from options in original zap pkg
```
OliverMKing, line 47: unnecessary comment. Perhaps rename rawOpts variable to something more fitting like fileLineOpts or defaultOpts.

OliverMKing (dismissed):

pkg/controller/controller.go
jaiveerk: oops deleted the wrong comment - will add the old one.
I'll call em default opts - just because in the future we may need to expand this to include more options, don't wanna limit it to one thing

davidgamero (approved):
lgtm

OliverMKing (approved):

PR #74, opened 2023-07-28
Add gocover checks for PR's

hsubramanianaks:
# Description

This PR is to add gocover and coverall- code coverage check for unit tests for any PR opened against main branch.

## Type of change: Feature

# How Has This Been Tested?

- [x] Github actions via forked repo

# Checklist:


.github/workflows/unit.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
         ref: ${{ inputs.ref }}
 
     - name: Set up Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
```
hsubramanianaks, line 21: This to fix a warning that we are using deprecated action, so updated this to v3.

coverage.out
```diff
@@ -0,0 +1,548 @@
+mode: set
```
hsubramanianaks, line 1: this file was generated from go test --cover-profile to get the coverage.out which is required for gocover.

.github/workflows/unit.yaml
```diff
@@ -35,3 +35,46 @@ jobs:
       if: ${{ !((github.event_name == 'repository_dispatch' && github.event.client_payload.slash_command.args.named.sha != '' && contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.args.named.sha)) || inputs.skipRefCheck) }}
       with:
         script: core.setFailed('Ref is not latest')
+  go-cover:
```
jaiveerk, line 38: I may be missing something but is there a reason why you made go-cover a separate job? couldn't you add the same steps to the previous job?

coverage.out
jaiveerk: can we gitignore this?

jaiveerk (changes requested):

coverage.out
bfoley13: I think we need to generate this file on the fly within the pipeline, otherwise we will have to update it constantly

OliverMKing: How will this work when `main` changes? Is there something automated that will update this in `main`?

OliverMKing: ah my comment is the same as brandon's

.github/workflows/unit.yaml
OliverMKing: i like separated too. Is there a reason this is under the unit workflow instead of a separate workflow entirely?

I remember gocover requiring unit tests to generate coverage at some point so that might be a reason but I haven't looked recently.

OliverMKing: but if gocover requires unit tests first I don't understand how this works since it's a separate job

.github/workflows/unit.yaml
```diff
@@ -35,3 +35,46 @@ jobs:
...
+      run: |
+        echo "download asset url from previous step is ${{ steps.get_release.outputs.result }}"
+        wget -q -O gocover.tar.gz ${{ steps.get_release.outputs.result }}
+        tar -xzf gocover.tar.gz
+        chmod +x gocover
+        echo $(pwd)
+        echo $(git branch)
+        echo $(git fetch --all)
+        ./gocover diff --cover-profile=coverage.out  --repository-path=./ --report-name=aks-app-routing-operator.report --outputdir=/tmp --compare-branch=origin/main  --coverage-baseline=80
+    - name: Upload coverage report
```
OliverMKing, line 74: It would be helpful if we could have a coverage report with coveralls like in the AgentBaker repo. https://github.com/Azure/AgentBaker/blob/d0a5f2a254a0209c927594484464c995c079555d/.github/workflows/check-coverage.yml#L38

It creates a nice comment on the PR breaking down where the coverage is for the PR. https://github.com/Azure/AgentBaker/pull/3471#issuecomment-1652335285

.github/workflows/unit.yaml
OliverMKing: Actually I see how this is working. It's taking the local coverage.out instead of regenerating. Same problem as here https://github.com/Azure/aks-app-routing-operator/pull/74/files#r1277680016

coverage.out
hsubramanianaks: Actually if we need separation then coverage.out needs to be in the repo and added to .gitignore. if not we can combine them together since go test runs before go cover which would generate coverage.out and can be used by gocover. so we need to decided which approach to follow. Please let me know your thoughts.

.github/workflows/unit.yaml
hsubramanianaks: https://github.com/Azure/aks-app-routing-operator/pull/74#discussion_r1277711868

coverage.out
OliverMKing: We should be generating the `coverage.out` by running go test on the pr 100%.

The big question is how we should be comparing to main. `main` branch needs its own `coverage.out` that we either automatically generate on the main branch or we attach to a workflow run and pull from the workflow run artifacts. I don't really like either of those solutions and would rather use Coveralls like agentbaker https://github.com/Azure/AgentBaker/blob/master/.github/workflows/check-coverage.yml.

hsubramanianaks: coverall requires workflow to have on push event, I don't know if it would work with repository dispatch_event like /ok-to-test. Please see screenshot below from coverall repo. 

image

Yes totally agree we can generate coverage.out but separation can't be achieved. Or we can duplicate go test with cover profile again in go cover stage/job.

.github/workflows/unit.yaml
hsubramanianaks: https://github.com/Azure/aks-app-routing-operator/pull/74#discussion_r1277727332

coverage.out
OliverMKing: There's no reason that unit tests can't be in a separate workflow than e2e tests. E2e is what needs repository_dispatch because that's the one that requires secrets.

OliverMKing: On the separation thing this would be part of the unit test workflow. You're right that there's no way to separate them.

It has to be part of the unit test job, though. Currently it's part of a separate job in the unit test workflow which doesn't work because jobs are different vms.

hsubramanianaks: @OliverMKing so it is totally okay to run unit tests on push events and separate it from e2e workflows ? Yes unit test and gocover has to run together so that coverage.out can be generated at runtime.

.github/workflows/unit.yaml
```diff
@@ -1,37 +1,27 @@
...
+      run: go test -race -v ./... -coverprofile=coverage.out
+
+    - name: Convert coverage to lcov
+      uses: jandelgado/gcov2lcov-action@v1
 
-    - name: Ensure ref
-      uses: actions/github-script@v6
-      if: ${{ !((github.event_name == 'repository_dispatch' && github.event.client_payload.slash_command.args.named.sha != '' && contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.args.named.sha)) || inputs.skipRefCheck) }}
+    - name: Coveralls
+      uses: coverallsapp/github-action@v2
```
OliverMKing, line 23: can we refer to this action by sha? More secure which is good for all 3rd party actions

.github/workflows/unit.yaml
```diff
@@ -1,37 +1,27 @@
...
-      if:
-        (github.event_name == 'repository_dispatch' &&
-        github.event.client_payload.slash_command.args.named.sha != '' &&
-        contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.args.named.sha)) ||
-        inputs.skipRefCheck
-      run: go test -race -v ./...
+      run: go test -race -v ./... -coverprofile=coverage.out
+
+    - name: Convert coverage to lcov
+      uses: jandelgado/gcov2lcov-action@v1
```
OliverMKing, line 20: can we refer to this action by sha? More secure which is good for all 3rd party actions

.github/workflows/unit.yaml
```diff
@@ -1,37 +1,27 @@
...
 
-on:
-  workflow_call:
-    inputs:
-      ref:
-        type: string
-      skipRefCheck:
-        type: boolean
-        default: true
+on: ["pull_request"]
```
OliverMKing, line 3: we should scope this down to main only since that's what we care about.

  push:
    branches: ["master"]
  pull_request:
    branches: ["master"]

also need the push: master since that's in the coveralls spec

OliverMKing: since we are changing this from a reusable workflow we need to remove this workflow from the other places it's called. 

https://github.com/Azure/aks-app-routing-operator/blob/main/.github/workflows/pr-validate-fork.yaml
https://github.com/Azure/aks-app-routing-operator/blob/main/.github/workflows/pr-validate.yaml

coverage.out
hsubramanianaks: We decided to use coverall in par with other OSS repo's like agent baker etc..

.github/workflows/unit.yaml
hsubramanianaks: Done replaced with coverall.

.github/workflows/unit.yaml
hsubramanianaks: Done.

.github/workflows/unit.yaml
hsubramanianaks: updated, thanks for all comments and suggestions.

.github/workflows/unit.yaml
hsubramanianaks: done, updated. thanks.

OliverMKing (dismissed):
LGTM! Thanks for the good work. 

The badge could maybe be added to the readme.md in this PR but that's not a big deal.

OliverMKing (approved):

jaiveerk (approved):

bfoley13 (approved):

PR #76, opened 2023-08-04
Add metrics to controller loops

NickKeller:
# Description

Adds metric reporting to the controller runs

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

- [ ] Unit tests grab the metric values before and after loops are run to verify that the metrics were incremented as expected

# Checklist:

### DIFF
--- a/pkg/controller/common/cleaner.go
+++ b/pkg/controller/common/cleaner.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-multierror"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -87,23 +88,35 @@ func (c *cleaner) Start(ctx context.Context) error {
 }
 
 func (c *cleaner) Clean(ctx context.Context) error {
+	var err error
+	defer func() {
+		//placing this call inside a closure allows for result and err to be bound after tick executes
+		//this makes sure they have the proper value
+		//just calling defer metrics.HandleControllerReconcileMetrics(controllerName, result, err) would bind
+		//the values of result and err to their zero values, since they were just instantiated
+		metrics.HandleControllerReconcileMetrics(c.controllerName(), ctrl.Result{}, err)
+	}()
 	if c.retriever == nil {
-		return errors.New("retriever is nil")
+		err = errors.New("retriever is nil")
+		return err
 	}
 
 	types, err := c.retriever(c.mapper)
 	if err != nil {
-		return fmt.Errorf("retrieving gvr types: %w", err)
+		err = fmt.Errorf("retrieving gvr types: %w", err)
+		return err
 	}
 
 	var result *multierror.Error
 	for _, t := range types {
-		if err := c.CleanType(ctx, t); err != nil {
-			result = multierror.Append(result, fmt.Errorf("cleaning type %s with labels %s: %w", t.gvr.String(), t.labels, err))
+		if cleanTypeErr := c.CleanType(ctx, t); cleanTypeErr != nil {
+			result = multierror.Append(result, fmt.Errorf("cleaning type %s with labels %s: %w", t.gvr.String(), t.labels, cleanTypeErr))
 		}
 	}
 
-	return result.ErrorOrNil()
+	err = result.ErrorOrNil()
+	return err
+
 }
 
 func (c *cleaner) CleanType(ctx context.Context, t cleanType) error {
@@ -166,6 +179,10 @@ func (c *cleaner) CleanType(ctx context.Context, t cleanType) error {
 
 func (c *cleaner) NeedLeaderElection() bool {
 	return true
+}
+
+func (c *cleaner) controllerName() string {
+	return fmt.Sprintf("%s_cleaner", c.name)
 }
 
 func isNamespaced(clientset kubernetes.Interface, gvr schema.GroupVersionResource) (bool, error) {
--- a/pkg/controller/common/cleaner.go
+++ b/pkg/controller/common/cleaner.go
@@ -184,7 +184,7 @@ func (c *cleaner) NeedLeaderElection() bool {
 }
 
 func (c *cleaner) controllerName() string {
-	return fmt.Sprintf("%s_cleaner", c.name)
+	return c.name
 }
 
 func isNamespaced(clientset kubernetes.Interface, gvr schema.GroupVersionResource) (bool, error) {
--- a/pkg/controller/common/cleaner.go
+++ b/pkg/controller/common/cleaner.go
@@ -58,6 +58,8 @@ func NewCleaner(manager ctrl.Manager, name string, gvrRetriever CleanTypeRetriev
 		retriever:  gvrRetriever,
 		maxRetries: 2,
 	}
+
+	metrics.InitControllerMetrics(c.controllerName())
 	return manager.Add(c)
 }
 
--- a/pkg/controller/common/resource_reconciler_test.go
+++ b/pkg/controller/common/resource_reconciler_test.go
@@ -2,6 +2,8 @@ package common
 
 import (
 	"context"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/testutils"
 	"testing"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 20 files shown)


pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -46,19 +51,28 @@ func NewPlaceholderPodController(manager ctrl.Manager, conf *config.Config, ingr
...
+		//the values of result and err to their zero values, since they were just instantiated
+		metrics.HandleControllerReconcileMetrics(placeholderPodControllerName, result, err)
+	}()
+
 	logger, err := logr.FromContext(ctx)
 	if err != nil {
-		return ctrl.Result{}, err
+		return result, err
 	}
 	logger = logger.WithName("placeholderPodController")
```
OliverMKing, line 70: The logger name can be derived from the controller name so if we ever need to correlate through some kind of query we can. This is beneficial because we add extra details to ICMs created by these metrics that will query for controller logs (filtered through logger name).

Maybe the logger name for each controller is the controller name just shifted from snake case to camel case? We  currently have no contract with anything in terms of logger names so the logger names could just be snake case instead.

OliverMKing: If that's a lot of work we can make a different story to do that

pkg/controller/common/cleaner.go
```diff
@@ -168,6 +183,10 @@ func (c *cleaner) NeedLeaderElection() bool {
 	return true
 }
 
+func (c *cleaner) controllerName() string {
+	return fmt.Sprintf("%s_cleaner", c.name)
```
OliverMKing, line 187: the one place the cleaner is used already does append the cleaner suffix already. We can figure out a cleaner way of handling controller names in a subsequent pr, just documenting
https://github.com/Azure/aks-app-routing-operator/blob/1028b62d138483de6dfca64c54ed39c11c54c50f/pkg/controller/dns/external_dns.go#L46C37-L46C55

pkg/controller/common/cleaner.go
```diff
@@ -57,6 +58,8 @@ func NewCleaner(manager ctrl.Manager, name string, gvrRetriever CleanTypeRetriev
 		retriever:  gvrRetriever,
 		maxRetries: 2,
 	}
+
+	metrics.InitControllerMetrics(c.controllerName())
```
OliverMKing, line 62: Any reason for not adding unit tests for the cleaner metrics?

OliverMKing: The Clean function itself isn't tested right now so that's probably why.

NickKeller: Yep exactly. Funny enough - the `Start` method also isn't called anywhere in the project? Not that Goland can tell anyways

pkg/controller/keyvault/event_mirror.go
```diff
@@ -44,54 +49,57 @@ func NewEventMirror(manager ctrl.Manager, conf *config.Config) error {
...
 	pod := &corev1.Pod{}
 	pod.Name = event.InvolvedObject.Name
 	pod.Namespace = event.InvolvedObject.Namespace
 	err = e.client.Get(ctx, client.ObjectKeyFromObject(pod), pod)
-	if errors.IsNotFound(err) {
-		return ctrl.Result{}, nil
-	}
 	if err != nil {
-		return ctrl.Result{}, err
+		return result, client.IgnoreNotFound(err)
```
OliverMKing, line 90: didn't know about client.IgnoreNotFound! Pretty cool!

pkg/controller/common/cleaner.go
NickKeller: In the interest of not being redundant - changed

pkg/controller/keyvault/event_mirror.go
NickKeller: There's also client.IgnoreAlreadyExists

pkg/controller/common/cleaner.go
```diff
@@ -87,23 +90,35 @@ func (c *cleaner) Start(ctx context.Context) error {
 }
 
 func (c *cleaner) Clean(ctx context.Context) error {
+	var err error
+	defer func() {
+		//placing this call inside a closure allows for result and err to be bound after tick executes
```
OliverMKing, line 95: are named returns a cleaner way of handling this rather than declaring the values before and always having to change a specific variable?

Named returns should make it so you don't have to always set err = something before returning. Would make it a lot harder to mess up.

If you return a value for a named return it sets that name to it so it will be used in the defer.

pkg/controller/service/ingress_reconciler.go
```diff
@@ -16,9 +16,14 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
```
OliverMKing, line 16: unit tests for this one?

OliverMKing (commented):
Left a few small nits. Really nice PR! Exactly what we need

pkg/controller/common/cleaner.go
OliverMKing: Start is done automatically by the controller manager. https://github.com/Azure/aks-app-routing-operator/blob/4af5d634ccab2fc09239b252a37a7e8766ee4741/pkg/controller/common/cleaner.go#L63

Start fulfills an interface needed there

pkg/controller/service/ingress_reconciler.go
NickKeller: added on next push

pkg/controller/common/cleaner.go
NickKeller: If we didn't have cases where we ignore errors and just return nil - yes. We have lots of cases where we return `ctrl.Result, client.IgnoreNotFound(err)`. In that case, we'd have to call `err = client.IgnoreNotFound(err)`, which can be just as awkward.

Ideally, this would be part of a pipeline, and whether or not an error is returned will determine if the error metrics are incremented - but that's work for a different PR

OliverMKing (approved):

pkg/controller/common/resource_reconciler_test.go
```diff
@@ -2,6 +2,8 @@ package common
 
 import (
 	"context"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/testutils"
```
jaiveerk, line 6: nit: imports

pkg/controller/ingress/concurrency_watchdog.go
```diff
@@ -135,11 +141,17 @@ func (c *ConcurrencyWatchdog) Start(ctx context.Context) error {
 
 func (c *ConcurrencyWatchdog) tick(ctx context.Context) error {
 	start := time.Now()
+	var retErr *multierror.Error
 	defer func() {
 		c.logger.Info("finished checking on ingress controller pods", "latencySec", time.Since(start).Seconds())
+
+		//placing this call inside a closure allows for result and err to be bound after tick executes
```
jaiveerk, line 148: I wonder if we should maybe just include this comment once in a README or something similar in the controller dir - that way we don't have to repeat it so frequently. 

open to leaving it though if we think that's the cleanest solution, just not a fan of how many times we repeat it

pkg/controller/ingress/concurrency_watchdog_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"container/ring"
 	"context"
 	"fmt"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/testutils"
```
jaiveerk, line 11: nit: imports

pkg/controller/keyvault/event_mirror_test.go
```diff
@@ -5,6 +5,8 @@ package keyvault
 
 import (
 	"context"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/testutils"
```
jaiveerk, line 9: nit: imports

pkg/controller/keyvault/ingress_secret_provider_class_test.go
```diff
@@ -6,6 +6,7 @@ package keyvault
 import (
 	"context"
 	"fmt"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/testutils"
```
jaiveerk, line 9: imports

pkg/controller/keyvault/placeholder_pod_test.go
```diff
@@ -5,6 +5,8 @@ package keyvault
 
 import (
 	"context"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
```
jaiveerk, line 8: same

pkg/controller/osm/ingress_cert_config_reconciler_test.go
```diff
@@ -5,6 +5,8 @@ package osm
 
 import (
 	"context"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
```
jaiveerk, line 8: here too

pkg/controller/service/ingress_reconciler_test.go
```diff
@@ -5,6 +5,8 @@ package service
 
 import (
 	"context"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/testutils"
```
jaiveerk, line 9: imports

jaiveerk (approved):
looks good, had some nits

OliverMKing:
Discussed nits with Jaiveer. We decided they aren't blocking so merged this in.

PR #77, opened 2023-08-04
Added unit tests

rsamigullin:
# Description

Improved code coverage for aks-app-routing-operator

Fixes # https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/24677451/
Feature # (details)

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes - Added more unit tests
Provide instructions so we can reproduce - run go test./... 

# Checklist:


jaiveerk (dismissed):
minor changes, lgtm otherwise

pkg/controller/controller_test.go
```diff
@@ -75,3 +76,11 @@ func TestGetSelfDeploy(t *testing.T) {
 		require.Nil(t, self)
 	})
 }
+
+func TestNewManagerForRestConfig(t *testing.T) {
+	testenv := &envtest.Environment{}
+	restConf, _ := testenv.Start()
+	conf := &config.Config{NS: "app-routing-system2", OperatorDeployment: "operator-test2", MetricsAddr: "0"}
```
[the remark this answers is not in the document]
rsamigullin: I had another unit test using `app-routing-system` name, but I ended up removing that unit test. Updated `TestNewManagerForRestConfig` and removed `2`s

pkg/controller/keyvault/event_mirror_test.go
```diff
@@ -113,3 +116,17 @@ func TestEventMirrorServiceOwnerHappyPath(t *testing.T) {
...
 }
+
+func TestNewEventMirror(t *testing.T) {
+	m := getManager()
+	conf := &config.Config{NS: "app-routing-system", OperatorDeployment: "operator"}
+	err := NewEventMirror(m, conf)
+	require.NoError(t, err, "should not error")
+}
+
+func getManager() manager.Manager {
```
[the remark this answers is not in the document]
rsamigullin: @jaiveerk great idea, I'll wait on Nick's pr then

pkg/controller/common/cleaner_test.go
```diff
@@ -237,3 +239,110 @@ func AssertAction(t *testing.T, got []k8stesting.Action, expected k8stesting.Act
...
+		err := test.c.Start(context.Background())
+		if err != nil {
+			require.Equal(t, test.expectedErrMsg, err.Error(), "should return expected error msg")
+		}
+	}
+}
+
+func getManager() manager.Manager {
+	testenv := &envtest.Environment{}
+	cfg, _ := testenv.Start()
```
OliverMKing, line 345: What is the result of never calling `testenv.End()`? I'm curious if that leaves a k8s environment hanging or if they are automatically cleaned at the end of the Go process?

I also wonder if we need a separate environment for each test or if it should be a global one. Most of the examples i see are global ones.

rsamigullin: I checked what `testenv.Stop()` does:
`Stop stops a running server. Previously installed CRDs, as listed in CRDInstallOptions.CRDs, will be uninstalled if CRDInstallOptions.CleanUpAfterUse are set to true` So not calling that function leaves env hanging. We should stick to global env.

.github/workflows/unit.yaml
```diff
@@ -17,6 +17,21 @@ jobs:
       with:
         go-version: ~1.20.0
 
+    - name: Create kubebuilder directory
```
OliverMKing, line 20: Why is this all being installed to a kubebuilder directory?

If it's a requirement we should have a comment explaining why because App Routing doesn't actually use the kubebuilder framework.

pkg/controller/dns/external_dns_test.go
```diff
@@ -52,10 +57,30 @@ var (
...
+
+	exitCode := m.Run()
+
+	testutils.CleanupTestingEnv()
+	os.Exit(exitCode)
+}
+
 var (
-	self *appsv1.Deployment = nil
+	restConfig *rest.Config
```
jaiveerk, line 82: nit: can we combine this with other var block or at least put it above testmain

jaiveerk (changes requested):

pkg/controller/common/cleaner_test.go
```diff
@@ -237,3 +238,109 @@ func AssertAction(t *testing.T, got []k8stesting.Action, expected k8stesting.Act
 
 	require.True(t, found, "expected to find action", expected)
 }
+
+func TestNewCleaner(t *testing.T) {
+	m, err := manager.New(restConfig, manager.Options{MetricsBindAddress: "0"})
```
OliverMKing, line 243: what's the reason for binding the metrics to 0?

pkg/controller/common/resource_reconciler_test.go
```diff
@@ -2,20 +2,34 @@ package common
...
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+var restConfig *rest.Config
+
+func TestMain(m *testing.M) {
+	restConfig, _ = testutils.StartTestingEnv()
```
OliverMKing, line 26: we should panic if we fail to start the testing env

pkg/controller/common/resource_reconciler_test.go
```diff
@@ -2,20 +2,34 @@ package common
...
+
+	exitCode := m.Run()
```
OliverMKing, line 28: moving the cleanup to a defer right under the start might make sense.

```Go
restCfg, err := testutils.StartTestingEnv()
if err != nil {
   panic(err)
}
defer testutils.CleanupTestingEnv()

os.Exist(m.Run())
```

pkg/controller/common/resource_reconciler_test.go
```diff
@@ -2,20 +2,34 @@ package common
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+var restConfig *rest.Config
+
+func TestMain(m *testing.M) {
```
OliverMKing, line 25: Maybe the TestMain should be moved to a more generic file within this package since it's not strictly tied to the resource_reconciler. `main_test.go` or something similar

pkg/controller/testutils/testutils.go
```diff
@@ -32,3 +37,12 @@ func GetReconcileMetricCount(t *testing.T, controllerName, label string) float64
 	beforeCount := metricProto.GetCounter().GetValue()
 	return beforeCount
 }
+
+func StartTestingEnv() (*rest.Config, error) {
+	env = &envtest.Environment{}
+	return env.Start()
+}
+
+func CleanupTestingEnv() error {
```
OliverMKing, line 46: We are using the same env cross package. Some tests start the environment which I believe is idempotent but the stop condition isn't. When one package calls stop it will teardown the environment for all tests still running. Do our tests not run in parallel? I'm curious about why this works and why we don't need a separate Environment for each package.

pkg/controller/common/cleaner_test.go
rsamigullin: `MetricsBindAddress` is the TCP address that the controller should bind to for serving prometheus metrics. It can be set to "0" to disable the metrics serving. I wanted to disable it because I didn't see a need for metrics and also for each `manager.New(...)` I would have to use different port in address, otherwise, I would get error: `error listening on :8080: listen tcp :8080: bind: address already in use`.

pkg/controller/dns/external_dns_test.go
rsamigullin: Done

pkg/controller/common/resource_reconciler_test.go
rsamigullin: Done

pkg/controller/common/resource_reconciler_test.go
rsamigullin: Added panic and defer

pkg/controller/common/resource_reconciler_test.go
rsamigullin: Moved to `main_test.go`

jaiveerk (dismissed):
approving to unblock merging

pkg/controller/testutils/testutils.go
rsamigullin: From my understanding, I don't think unit tests run in parallel, especially from different modules. In each module testenv Starts, runs unit tests withing module, Stops. Here is example when I run it locally with added `STARTING, STOPPING` logs:
image

We can run tests in parallel within single module by using -parallel.
Also, does go run unit tests concurrently ? Looking into that I'm not sure
https://github.com/golang/go/blob/master/src/testing/testing.go#L2025

.github/workflows/unit.yaml
OliverMKing: It seems fine to keep as is just documenting the reason with a comment would be good.

pkg/controller/common/cleaner_test.go
OliverMKing: makes sense!

pkg/controller/testutils/testutils.go
OliverMKing: https://coderwall.com/p/rwgo_a/go-test-package-in-parallel

My understanding is that tests within a package are run sequentially by default but all packages are run in parallel.

rsamigullin: I found out that each `TestMain` function creates NEW in-memory testing environment with different `restConfig.Host` values.
I think that's why unit tests are not failing.

.github/workflows/unit.yaml
rsamigullin: Added comment for GH action

pkg/controller/testutils/testutils.go
```diff
@@ -32,3 +37,12 @@ func GetReconcileMetricCount(t *testing.T, controllerName, label string) float64
 	beforeCount := metricProto.GetCounter().GetValue()
 	return beforeCount
 }
+
+func StartTestingEnv() (*rest.Config, error) {
+	env = &envtest.Environment{}
```
OliverMKing, line 42: This is assigning env to a new envtest environment every time it's called. How do old ones get cleaned up?

OliverMKing: I'm not sure how much value having a utils for the testing env is bringing. Now we have to solve the problem of keeping state of each different test package env all in the same place? Doesn't seem worth it over just defining a envtest.Environment and starting and stopping in each package.

OliverMKing: Thought about it some more. We can have a singleton of testenv

```Go
var env *envtest.Environment
var usages int

func EnsureTestingEnv() (*rest.Config, error) {
   if env != nil {
       usages++
       return env.GetConfig(), nil // don't actually know the fn for getting the config for an env
   }

   new := &envtest.Environment{}
   cfg, err := env.Start()
   if err != nil {
      return nil, fmt.Errorf("starting environment: %w", err)
   }
   
   env = new
   usages++
   return cfg, nil
}

func ReleaseTestingEnv() error {
    usages--
    if usages < 1 {
        return env.Stop()
    }
   
    return nil
}
```

This would be the general strategy but we'd need to use either the atomic or sync packages to make this thread safe.

OliverMKing: This could exist in utils and solve all the problems.

rsamigullin: Solved by added envtest variable to start/stop functions:
```
func StartTestingEnv() (*rest.Config, *envtest.Environment, error) {
	env = &envtest.Environment{}
	restConfig, err := env.Start()
	if err != nil {
		return nil, nil, err
	}
	return restConfig, env, nil
}

func CleanupTestingEnv(env *envtest.Environment) error {
	return env.Stop()
}
```

pkg/controller/testutils/testutils.go
```diff
@@ -1,12 +1,17 @@
...
+
 	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
 	promDTO "github.com/prometheus/client_model/go"
 	"github.com/stretchr/testify/require"
-	"testing"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
 )
 
+var env *envtest.Environment
```
OliverMKing, line 13: we shouldn't have a global anymore since env is returned by start

pkg/controller/testutils/testutils.go
```diff
@@ -32,3 +37,16 @@ func GetReconcileMetricCount(t *testing.T, controllerName, label string) float64
 	beforeCount := metricProto.GetCounter().GetValue()
 	return beforeCount
 }
+
+func StartTestingEnv() (*rest.Config, *envtest.Environment, error) {
+	env = &envtest.Environment{}
```
OliverMKing, line 42: slight race condition of still using the global variable.

Two tests can be in startTestingEnv at the same time and overwrite eachother's env.

OliverMKing (approved):

bfoley13 (approved):

PR #80, opened 2023-08-14
Match prometheus controller labels to controller name 

Marcus-Hines:
…in line with prometheus best practices

# Description
Match prometheus controller labels to controller name - Adding the function to check for snakeCase and its unit test first. Next step is documenting that the controller name needs to be snake case, and updating the the loggers to use the correct controller name / label

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -6,6 +6,7 @@ package controller
 import (
 	"context"
 	"net/http"
+	"regexp"
 
 	"github.com/Azure/aks-app-routing-operator/pkg/controller/dns"
 	"github.com/Azure/aks-app-routing-operator/pkg/controller/ingress"
@@ -144,6 +145,15 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
 	}
 
 	return m, nil
+}
+
+// IsPrometheusBestPracticeName - function returns true if the name given matches best practices for prometheus name
+func IsPrometheusBestPracticeName(controllerName string) bool {
+	pattern := "[a-z]+(_[a-z]+)*"
+	re := regexp.MustCompile(pattern)
+	match := re.FindStringSubmatch(controllerName)
+
+	return match != nil && match[0] == controllerName
 }
 
 func getSelfDeploy(kcs kubernetes.Interface, conf *config.Config, log logr.Logger) (*appsv1.Deployment, error) {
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -146,6 +147,15 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
 	return m, nil
 }
 
+// IsPrometheusBestPracticeName - function returns true if the name given matches best practices for prometheus name, i.e. snake_case
```
OliverMKing, line 150: Looks pretty good to me. Should be moved to testutils https://github.com/Azure/aks-app-routing-operator/blob/main/pkg/controller/testutils/testutils.go

since this will only be a unit testing check

pkg/controller/controller.go
```diff
@@ -146,6 +147,15 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
...
+func IsPrometheusBestPracticeName(controllerName string) bool {
+	pattern := "[a-z]+(_[a-z]+)*"
+	re := regexp.MustCompile(pattern)
+	match := re.FindStringSubmatch(controllerName)
```
OliverMKing, line 154: This would be a bit cleaner if we use

https://pkg.go.dev/regexp#Regexp.MatchString

`MatchString` and control for the entire string with ^ and $ regex.

pkg/controller/controller.go
Marcus-Hines: I understand your reasoning. But let me clarify. So it's only a unit testing check at the moment -  only because I like to introduce functionality and its unit test into the code base, before actually using it in throughout the code. But, it will be a function called by every controller before initializing the logger,  so I think its reasonable to put it here. The next PR you will see it called by `cleaner.go`

pkg/controller/controller_test.go
```diff
@@ -103,3 +103,18 @@ func TestNewManagerForRestConfig(t *testing.T) {
 	_, err := NewManagerForRestConfig(conf, restConfig)
 	require.NoError(t, err)
 }
+
+func TestIsPrometheusBestPracticeName(t *testing.T) {
```
OliverMKing, line 107: this test should likely be moved into a `testutils_test.go` file

OliverMKing (approved):

bfoley13 (approved):

jaiveerk (approved):

PR #85, opened 2023-08-24
add controller name interface to all controllers

OliverMKing:
# Description

Adds controller name interface to all controllers. Allows for us to enforce consistent logging and metric naming for all controllers. Will be useful for alerting based on prometheus especially. If an alert fires we can attach an automatically generated query that can search relevant controller logs for that alert.

## Type of change

# How Has This Been Tested?

Unit and e2e tested

# Checklist:

### DIFF
--- a/pkg/controller/common/cleaner.go
+++ b/pkg/controller/common/cleaner.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/Azure/aks-app-routing-operator/pkg/controller/controllername"
 	"github.com/Azure/aks-app-routing-operator/pkg/controller/metrics"
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-multierror"
@@ -23,7 +24,7 @@ import (
 )
 
 type cleaner struct {
-	name       string
+	name       controllername.ControllerNamer
 	mapper     meta.RESTMapper
 	clientset  kubernetes.Interface
 	dynamic    dynamic.Interface
@@ -33,7 +34,7 @@ type cleaner struct {
 }
 
 // NewCleaner creates a cleaner that attempts to delete resources with the labels specified and of the types returned by CleanTypeRetriever
-func NewCleaner(manager ctrl.Manager, name string, gvrRetriever CleanTypeRetriever) error {
+func NewCleaner(manager ctrl.Manager, name controllername.ControllerNamer, gvrRetriever CleanTypeRetriever) error {
 	d, err := dynamic.NewForConfig(manager.GetConfig())
 	if err != nil {
 		return fmt.Errorf("creating dynamic client: %w", err)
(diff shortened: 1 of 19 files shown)


pkg/controller/controllername/controllername.go
```diff
@@ -3,37 +3,49 @@ package controllername
...
 type controllerName []string
 
-func NewControllerName(name []string) controllerName {
-	cn := make(controllerName, len(name))
-
-	for i, w := range name {
-		cn[i] = strip(strings.ToLower(w))
-
+// New returns a new controllerName after taking each word of the controller name as a separate argument
+func New(firstWord string, additionalWords ...string) controllerName { // using a non-variadic for the first word makes it impossible to accidentally pass no arguments in. Accepting variadic versus slices also helps with not accepting empty slices and is easier to use
```
davidgamero, line 31: what happens if the first word is an empty string and there's no variadic inputs?

pkg/controller/common/cleaner.go
```diff
@@ -23,7 +24,7 @@ import (
 )
 
 type cleaner struct {
-	name       string
+	name       controllername.ControllerNamer
```
davidgamero, line 27: nit if the `name` field is a `namer` then does the name:`namer` return also a name? idk if changing this to a "namer" field is clearer or makes it more complex

Marcus-Hines (approved):
LGTM

jaiveerk (approved):
very clean, lgtm

pkg/controller/controllername/controllername.go
OliverMKing: then the controller name is just an empty string which isn't that bad. The metrics label for the controller will be "" and the logger fields will be "" which doesn't break anything. 

We try to avoid that as much as possible with the entire args not being variadic which means someone has to call this like
```Go
controllername.New("")

```
which should obviously look wrong to any user and should be caught during any code review hopefully

jaiveerk: I think that would result in a blank name but feel like that's a pretty niche use-case? I guess we could add an error to the return and error if it's blank?

OliverMKing: this is as good as it's going to get without a linter (overkill for this) or runtime verification (can't have because we shouldn't crash operator due to a bad name). We can't fully protect against a consumer consuming this package completely wrong but that's pretty much true of every single package, there's only so much you can do.

pkg/controller/common/cleaner.go
jaiveerk: we could rename the field to controllername? or controllernamer? I think I'd prefer it be called "namer" rather than name since it's something that generates names for different situations

pkg/controller/controllername/controllername.go
davidgamero: sounds good. i agree this provides the most protection

davidgamero (approved):
lgtm

pkg/controller/common/cleaner.go
OliverMKing: going to resolve so i can merge and get a release out. namer is maybe a better name and can rename in a future pr after we hit this release deadline

PR #90, opened 2023-08-29
Add new e2e testing framework

OliverMKing:
# Description

Fixes how the operator looks for the self deploy by adding a new field for that.

Finishes another phase of the e2e testing framework. This framework isn't used yet but will be fully migrated to in the future.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Tested locally

# Checklist:


pkg/config/config.go
```diff
@@ -37,6 +37,7 @@ func init() {
 	flag.StringVar(&Flags.ServiceAccountTokenPath, "service-account-token-path", "", "optionally override the default token path")
 	flag.StringVar(&Flags.MetricsAddr, "metrics-addr", "0.0.0.0:8081", "address to serve Prometheus metrics on")
 	flag.StringVar(&Flags.ProbeAddr, "probe-addr", "0.0.0.0:8080", "address to serve readiness/liveness probes on")
+	flag.StringVar(&Flags.OperatorNS, "operator-namespace", "kube-system", "namespace of the operator's k8s deployment")
```
jaiveerk, line 40: can we add a UT for this just to maintain coverage? also - if we maintain this, will the operator begin deploying to this ns in the overlay?

testing/e2e/clients/aks.go
```diff
@@ -119,43 +128,160 @@ func (a *aks) Deploy(ctx context.Context, objs []client.Object) error {
...
+					lgr.Info("waiting for job complete")
+					if err := a.runCommand(ctx, armcontainerservice.RunCommandRequest{
+						Command: to.Ptr(fmt.Sprintf("kubectl logs --pod-running-timeout=20s --follow job/%s -n %s", obj.GetName(), ns)),
+					}, runCommandOpts{
+						outputFile: fmt.Sprintf("job-%s.log", obj.GetName()), // output to a file for jobs because jobs are naturally different from other deployment resources in that waiting for "stability" is waiting for them to complete
+					}); err != nil {
+						return fmt.Errorf("waiting for job/%s to complete: %w", obj.GetName(), err)
+					}
+
+					lgr.Info("checking job statuss")
```
jaiveerk, line 242: nit: typo

testing/e2e/clients/aks.go
```diff
@@ -119,43 +128,160 @@ func (a *aks) Deploy(ctx context.Context, objs []client.Object) error {
...
+						Command: to.Ptr(fmt.Sprintf("kubectl wait --for=condition=Ready pod/%s -n %s", obj.GetName(), ns)),
+					}, runCommandOpts{}); err != nil {
+						return fmt.Errorf("waiting for pod/%s to be stable: %w", obj.GetName(), err)
+					}
+				case kind == "Job":
+					lgr.Info("waiting for job complete")
+					if err := a.runCommand(ctx, armcontainerservice.RunCommandRequest{
+						Command: to.Ptr(fmt.Sprintf("kubectl logs --pod-running-timeout=20s --follow job/%s -n %s", obj.GetName(), ns)),
+					}, runCommandOpts{
+						outputFile: fmt.Sprintf("job-%s.log", obj.GetName()), // output to a file for jobs because jobs are naturally different from other deployment resources in that waiting for "stability" is waiting for them to complete
```
jaiveerk, line 237: simply because jobs can be annoying about not being able to redeploy them without deleting the original one - do we want to include cleanup here once we know the job is done/once we've gotten its output? or just leave that to user/test dev?

jaiveerk: this would only really be useful in the event of a failed job/test so not sure if it's necessary/desirable to cater to that scenario

jaiveerk (changes requested):

testing/e2e/clients/aks.go
OliverMKing: We have a cleanup step already it's just not bundled with the stability checker since stability has a different goal. We attempt to cleanup any old jobs before deploying a new one.

pkg/config/config.go
jaiveerk: re: UT - understandable makes sense

so part of rolling out this next release will entail feeding in the ccp ns as a command line argument?
I guess I'm just confused on what the intention is with this arg vs the NS arg

OliverMKing: NS arg is where things like external dns and nginx are deployed to. OperatorNS is where the operator is located. The two are completely different right now (app-routing-namespace vs kube-system). We need OperatorNS so we can set ownership references which are essential to garbage collection

testing/e2e/clients/aks.go
jaiveerk: do you mean in deploy_e2e_tester.sh? will we still use that script when we transition to this go-based provisioning?

testing/e2e/suites/suites.go
jaiveerk, line 1: can we rename this to be called test.go since it defines the test type and its methods?

jaiveerk (changes requested):
just some feedback about file naming/placement but lgtm otherwise

pkg/controller/controller.go
```diff
@@ -147,9 +147,14 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
 }
 
 func getSelfDeploy(kcs kubernetes.Interface, conf *config.Config, log logr.Logger) (*appsv1.Deployment, error) {
-	deploy, err := kcs.AppsV1().Deployments(conf.NS).Get(context.Background(), conf.OperatorDeployment, metav1.GetOptions{})
+	ns := conf.OperatorNS
+	if ns == "" {
+		ns = "default"
```
bfoley13, line 152: why default here and kube-system default in the flag?

testing/e2e/tests/test.go
```diff
@@ -0,0 +1,67 @@
...
+		if err := t.Run(ctx); err != nil {
+			return logger.Error(lgr, err)
+		}
+
+		lgr.Info("finished running test")
+		return nil
+	}
+
+	lgr.Info("running sequential tests")
+	for _, test := range sequential {
```
bfoley13, line 44: Can we define tests in such a way the parallel jobs can be kicked off first with sequential jobs starting at the same time? Or is this an infra restriction? Can you provide an example of what youre thinking that would block it

OliverMKing: It's an infra restriction. Sequential tests would be tests that can't be run at the same time as any other test.

An example sequential test would be testing the cleanup of external dns when no dns ids are passed in to the operator (we remove any running external dns deployments). That shouldn't run at the same time as other tests because it could impact them since other tests would rely on external dns.

pkg/controller/controller.go
bfoley13: I suppose if were following the same logic as not providing a namespace means default im good with that

testing/e2e/tests/test.go
OliverMKing: Yes. No reason to provision multiple of the same cluster unless there's a cluster level configuration option.

We deploy a few different kinds of clusters (osm enabled, large node pool, other cluster-level configurations) then run all tests on those clusters (unless that test wouldn't work on that cluster through the `ShouldRun(infra infra.Provisioned) bool` function). An example of a test not working on a certain kind of infra would be something exclusively testing osm that wouldn't work on non-osm clusters.

tbarnes94 (approved):

jaiveerk (approved):

testing/e2e/tests/test.go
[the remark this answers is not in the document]
OliverMKing: will do in future. expect for this to change slightly when moving tests

testing/e2e/suites/suites.go
OliverMKing: will do in future. expect for this to change slightly when moving tests

PR #95, opened 2023-09-07
Move basic e2e tests to new e2e and add upgrade scenario

OliverMKing:
# Description

- Moves a few tests to the new testing framework.
- Adds framework for testing different operator versions and configurations
- Swaps around a few things that weren't working (like how we track things happening in private clusters with the aks invoke api). AKS invoke api isn't good for long running operations.

What's left for e2e

- Need to move a few more tests over (trivial since tests are fully working here)
- Need to setup GitHub Action workflow to run tests 
- Add a little more documentation to how to use it after GitHub workflow is done. Add notes to release steps for adding new releases here.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Tested locally. You can too by running

```sh
cd testing/e2e/
go run ./main.go infra --subscription=<subscription> --tenant=<tenant> --names="basic cluster"
go run ./main.go deploy
```

# Checklist:


jaiveerk (changes requested):
very much left to go through but wanted to post these before the weekend. 
excited to look through the rest!

testing/e2e/manifests/clientServer.go
```diff
@@ -0,0 +1,145 @@
...
+		setGroupKindVersion(obj)
+	}
+
+	return ret
+}
+
+func ClientAndServer(namespace, name, zoneName, nameserver, keyvaultURI string) testingResources {
+	host := strings.ToLower(namespace) + "." + zoneName
+
+	clientDeployment := newGoDeployment(clientContents, namespace, name+"-client")
```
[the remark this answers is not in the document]
OliverMKing: We cannot. We shouldn't mix things here and in RP because this is OSS. Also this repo is the source of any new code changes so this repo would be the source of truth (not RP if we wanted to do things that way).

testing/e2e/manifests/clientServer.go
```diff
@@ -0,0 +1,145 @@
...
+func (t testingResources) Objects() []client.Object {
+	ret := []client.Object{
+		t.Client,
+		t.Server,
+		t.Service,
+		t.Ingress,
+	}
+
+	for _, obj := range ret {
+		setGroupKindVersion(obj)
```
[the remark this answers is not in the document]
OliverMKing: obj is an interface. The underlying type is a pointer.

testing/e2e/clients/dns.go
```diff
@@ -73,11 +75,17 @@ func NewZone(ctx context.Context, subscriptionId, resourceGroup, name string, zo
 		return nil, fmt.Errorf("creating zone: %w", err)
 	}
 
+	nameservers := make([]string, len(resp.Properties.NameServers))
+	for i, ns := range resp.Properties.NameServers {
+		nameservers[i] = *ns
```
bfoley13, line 80: you are de-referencing a lot of variables without checks here and in the resp

testing/e2e/manifests/clientServer.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	serverName := name + "-server"
+	serverDeployment := newGoDeployment(serverContents, namespace, serverName)
+	serviceName := name + "service"
+	ingressName := name + "-ingress"
+
+	service :=
+		&corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{
```
bfoley13, line 84: you can move these service/ingress/etc. to the common file for reuse in the future

testing/e2e/manifests/ns.go
```diff
@@ -0,0 +1,18 @@
+package manifests
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UncollisionedNs returns a namespace with a guaranteed unique name after creating the namespace
+func UncollisionedNs() *corev1.Namespace {
```
bfoley13, line 9: move to common

testing/e2e/suites/utils.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		for _, condition := range d.Status.Conditions {
+			if condition.Type == appsv1.DeploymentAvailable && condition.Status == "True" {
+				lgr.Info("deployment is available")
+				return nil
+			}
+		}
+
+		if time.Since(start) > 10*time.Minute { // 6 minutes because it takes a decent amount of time for dns to "propagate"
```
bfoley13, line 31: nit: comment says 6 minutes

testing/e2e/suites/basic.go
```diff
@@ -0,0 +1,70 @@
...
+	publicBasicIngressNs = manifests.UncollisionedNs() // using a shared ns allows us to appropriately test upgrade scenarios
+)
+
+func basicSuite(infra infra.Provisioned) []test {
+	return []test{
+		{
+			name: "public basic ingress",
+			cfgs: builderFromInfra(infra).
+				withOsm(false).
+				withVersions(manifests.OperatorVersionLatest, manifests.OperatorVersion0_0_3).
```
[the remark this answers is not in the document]
OliverMKing: swapped

testing/e2e/suites/operatorConfig.go
```diff
@@ -0,0 +1,96 @@
...
+		osmEnabled: enabled,
+	}
+}
+
+type cfgBuilderWithVersions struct {
+	cfgBuilderWithOsm
+	versions []manifests.OperatorVersion
+}
+
+func (c cfgBuilderWithOsm) withVersions(versions ...manifests.OperatorVersion) cfgBuilderWithVersions {
```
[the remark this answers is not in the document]
OliverMKing: Added an all manifests variable to get all. We don't have any exclusions on versions so far. I'll add that function when we get to that.

Not sure withExclusions is the right way. When we need them we'll need exclusions because the operator will behave in different ways. Maybe instead of having inclusions we have subsets of All like

`AllOperatorVersions`, `OperatorVersionsWithXFeature`, etc

Will think about it more when we get there

testing/e2e/manifests/ns.go
OliverMKing: moved

testing/e2e/clients/dns.go
```diff
@@ -73,11 +75,17 @@ func NewZone(ctx context.Context, subscriptionId, resourceGroup, name string, zo
 		return nil, fmt.Errorf("creating zone: %w", err)
 	}
 
+	nameservers := make([]string, len(resp.Properties.NameServers))
+	for i, ns := range resp.Properties.NameServers {
```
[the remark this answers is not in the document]
OliverMKing: added them to guard against azure sdk breaking contracts

testing/e2e/manifests/clientServer.go
OliverMKing: Not sure how useful that is. Ingresses/services should pretty much always be bundled with the client server for our tests since that's how we actually test an ingress. I don't think we'll ever need a standalone one but if we did it would make sense to move to common. Any particular use you have in mind?

bfoley13: Not a standalone service/ingress, this would just make it easier for reuse if we need it later on, and clean up these definitions, like you have the deployment creation function

bfoley13 (approved):

docs/e2e.md
```diff
@@ -0,0 +1,67 @@
...
+
+You can also run E2E locally. See [Local E2E](#-local-e2e).
+
+## Architecture
+
+### Major Steps
+
+E2e is designed so that every major step is run in a separate GitHub Job. This allows for easy retries and makes it a lot more obvious about which step failed. It also means logs are easier to read.
+
+![e2e-jobs](./e2e-jobs.png)
```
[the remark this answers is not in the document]
OliverMKing: wil do in next PR

testing/e2e/manifests/clientServer.go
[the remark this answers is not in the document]
jaiveerk: that's a good point - wonder if it's worth putting client/server functionality in an exportable module and having RP import that

testing/e2e/suites/operatorConfig.go
```diff
@@ -0,0 +1,96 @@
...
+
+func builderFromInfra(infra infra.Provisioned) cfgBuilder {
+	return cfgBuilder{
+		msi:      infra.Cluster.GetClientId(),
+		tenantId: infra.TenantId,
+		location: infra.Cluster.GetLocation(),
+	}
+}
+
+type cfgBuilderWithOsm struct {
```
jaiveerk, line 22: out of curiosity - why implement this step-wise approach with new fields? why not just have one config struct with all the fields?

docs/e2e.md
```diff
@@ -0,0 +1,67 @@
...
+Each test defines a matrix of options it will run against including compatible operator versions and operator configs. Then when running tests a testing order is defined where each test with the same operator version and operator config run in parallel. Operator version is incremented appropriately to simulate an upgrade scenario.
+
+For example, we might have the following tests
+
+| Name   | Versions      | Public Dns Zones | Private Dns Zones |
+|--------|---------------|------------------|-------------------|
+| Test 1 | Latest, 0.0.3 | One, Multiple    | One               |
+| Test 2 | Latest, 0.0.3 | One              | One               |
+
+The testing order would be calculated as something like this.
```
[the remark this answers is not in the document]
jaiveerk: update now that I've read through everything I think just a line mentioning operatorConfig.go or just the build() function would be useful

testing/e2e/tests/types.go
```diff
@@ -3,25 +3,55 @@ package tests
...
+	GetOperatorConfigs() []manifests.OperatorConfig
+	test
 }
 
 // Ts is a slice of T
 type Ts []T
+
+type ordered []testsWithRunInfo
+
+type testsWithRunInfo struct {
```
jaiveerk, line 28: there are lots of types here and it's a bit confusing to keep track of all the casting/movement of information which I imagine will make debugging a bit difficult - is there any way we can clean this up? This was mostly inspired by all the different types used in testing/e2e/tests/run.go

jaiveerk: realistically I'm not sure how exactly this will take place but happy to chat about what my specific pain points are - just a lot of mixes between lists and types that contain lists, similar names, etc etc - I just feel like it can be cleaner/that we can consolidate a lot of the information that is repeatedly moving around

testing/e2e/tests/run.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	// combine tests that use the same operator configuration and operator version, so they can run in parallel
+	lgr.Info("grouping tests by operator configuration")
+	ret := make(ordered, 0)
+	for _, version := range versions {
+		// group tests by operator configuration
+		operatorCfgSet := make(map[manifests.OperatorConfig][]testWithConfig)
+		for _, test := range operatorVersionSet[version] {
+			operatorCfgSet[test.config] = append(operatorCfgSet[test.config], test)
```
jaiveerk, line 142: re: my comment in types... this is a lot

jaiveerk (changes requested):
lgtm for the most part, only request is if we can clean up test setup/running suites with different configs (even if that means, for instance, adding operator deploy scenario as an operator config or something else to consolidate information there)

testing/e2e/clients/aks.go
```diff
@@ -231,19 +273,50 @@ func (a *aks) waitStable(ctx context.Context, objs []client.Object) error {
...
+					if err := os.RemoveAll(outputFile); err != nil {       // clean out previous log file, if doesn't exist returns nil
+						return fmt.Errorf("removing previous job log file: %w", err)
 					}
 
-					lgr.Info("checking job status")
-					if err := a.runCommand(ctx, armcontainerservice.RunCommandRequest{
-						Command: to.Ptr(fmt.Sprintf("kubectl wait --for=condition=complete --timeout=10s job/%s -n %s", obj.GetName(), ns)),
-					}, runCommandOpts{}); err != nil {
-						return fmt.Errorf("waiting for job/%s to complete: %w", obj.GetName(), err)
+					getLogsFn := func() error { // right now this just dumps all logs on the pod, if we eventually have more logs
```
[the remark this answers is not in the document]
OliverMKing: this works for now

testing/e2e/manifests/operator.go
```diff
@@ -16,43 +19,162 @@ var (
...
+
+	// AllOperatorVersions is a list of all the operator versions
+	AllOperatorVersions = []OperatorVersion{OperatorVersion0_0_3, OperatorVersionLatest}
+)
+
+// OperatorVersion is an enum for the different versions of the operator
+type OperatorVersion uint
+
+const (
+	OperatorVersion0_0_3 OperatorVersion = iota // use iota to number with earlier versions being lower numbers
```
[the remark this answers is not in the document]
OliverMKing: this needs to be a constant. We might need to test more versions in the future.

testing/e2e/manifests/operator.go
```diff
@@ -16,43 +19,162 @@ var (
...
 		"--location", o.Location,
-		"--cluster-uid", o.ClusterUid,
+		"--namespace", "app-routing-system",
+		"--cluster-uid", "test-cluster-uid",
 	}
 
-	if len(o.DnsZoneIds) > 0 {
-		ret = append(ret, "--dns-zone-ids", strings.Join(o.DnsZoneIds, ","))
+	if o.Version == OperatorVersionLatest {
+		ret = append(ret, "--dns-sync-interval", (time.Second * 15).String())
```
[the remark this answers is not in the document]
OliverMKing: Yes. It's a tradeoff of cpu / mem usage but also api calls. There is throttling if there's too many calls to azure dns.

From a cx perspective there's no reason for a super fast sync of dns records. Once it's synced once it just works. If you're adding it you won't expect that it works immediately.

testing/e2e/suites/operatorConfig.go
OliverMKing: one config struct with all fields does not work well with multiple dimensions.

The point of this builder is to easily support multiple dimensions which most tells will feature. This builder also doesn't stop someone from using a single config struct.



Example:

Say I'm defining that it works with the configs

- Feature A on, Feature B off
- Feature A off, Feature B off
- Feature A on, Feature B on
- Feature A off, Feature B on

That takes 4 definitions and if we add a new "feature" that will take 8 definitions

Better to use the builder pattern where we can say

Builder.WithFeatureA(on, off).WithFeatureB(on, off) which outputs the same thing

testing/e2e/tests/types.go
OliverMKing: it's definitely a lot but using more times makes this simpler in this case. Using separate types to move information up means we can take away information making it easier to develop with.

If you can think of a specific way to improve this I'm happy to implement that in the future but this is the best I have for now.

testing/e2e/suites/operatorConfig.go
```diff
@@ -0,0 +1,96 @@
+package suites
+
+import (
+	"github.com/Azure/aks-app-routing-operator/testing/e2e/infra"
+	"github.com/Azure/aks-app-routing-operator/testing/e2e/manifests"
+)
+
+type cfgBuilder struct {
```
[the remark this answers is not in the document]
OliverMKing: what exactly are you looking to be documented here? Tried to make it as self documenting as possible with the type system

jaiveerk (approved):
approving changes per discussion/call

PR #100, opened 2023-09-21
add prometheus e2ev2 tests

OliverMKing:
# Description

Adds Prometheus tests and modifies current tests to make e2ev2 have all tests that e2ev1 has.

## Type of change

# How Has This Been Tested?

This is the test. Ran e2e and it all works.

# Checklist:

### DIFF
--- a/.github/workflows/e2ev2-provision-test.yaml
+++ b/.github/workflows/e2ev2-provision-test.yaml
@@ -92,7 +92,7 @@ jobs:
       - name: Print test logs
         shell: bash
         run: cat testing/e2e/job-app-routing-operator-e2e.log
-        if: always() && steps.test.outcome == 'failure'
+        if: always() && (steps.test.outcome == 'failure' || steps.test.outcome == 'success')
 
       - name: Ensure ref
         uses: actions/github-script@v6
--- a/testing/e2e/infra/infras.go
+++ b/testing/e2e/infra/infras.go
@@ -2,12 +2,19 @@ package infra
 
 import (
 	"github.com/Azure/aks-app-routing-operator/testing/e2e/clients"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v2"
 	"github.com/google/uuid"
 )
 
 var (
 	rg       = "app-routing-e2e" + uuid.New().String()
 	location = "South Central US"
+)
+
+var (
+	// OsmInfraName is the name of the infrastructure that will be used to test an OSM enable cluster
+	OsmInfraName = "osm cluster"
 )
 
 // Infras is a list of infrastructure configurations the e2e tests will run against
@@ -24,6 +31,25 @@ var Infras = infras{
 		Location:      location,
 		Suffix:        uuid.New().String(),
 		McOpts:        []clients.McOpt{clients.PrivateClusterOpt},
+	},
+	{
+		Name:          OsmInfraName,
+		ResourceGroup: rg,
+		Location:      location,
+		Suffix:        uuid.New().String(),
+		McOpts: []clients.McOpt{
+			func(mc *armcontainerservice.ManagedCluster) error {
+				if mc.Properties.AddonProfiles == nil {
+					mc.Properties.AddonProfiles = map[string]*armcontainerservice.ManagedClusterAddonProfile{}
+				}
+
+				mc.Properties.AddonProfiles["openServiceMesh"] = &armcontainerservice.ManagedClusterAddonProfile{
+					Enabled: to.Ptr(true),
+				}
+
+				return nil
+			},
+		},
 	},
 }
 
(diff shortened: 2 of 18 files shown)


.github/workflows/e2ev2-provision-test.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Print test logs
         shell: bash
         run: cat testing/e2e/job-app-routing-operator-e2e.log
-        if: always() && steps.test.outcome == 'failure'
+        if: always() && (steps.test.outcome == 'failure' || steps.test.outcome == 'success')
```
bfoley13, line 95: What are all the options for the outcome?

testing/e2e/manifests/common.go
```diff
@@ -61,6 +64,30 @@ func setGroupKindVersion(obj client.Object) {
...
+	f, err := parser.ParseFile(fset, "", contents, parser.ImportsOnly)
+	if err != nil {
+		panic(fmt.Errorf("parsing contents: %w", err)) // this should only happen if the contents are invalid Go code
+	}
+	for _, imp := range f.Imports {
+		if strings.Contains(imp.Path.Value, ".") { // if it contains a dot it's not a standard library import
+			command = []string{
+				"/bin/sh",
+				"-c",
+				"mkdir source && cd source && go mod init source && echo '" + contents + "' > main.go && go mod tidy && go run main.go",
```
bfoley13, line 85: any reason we dont just use this the command definition above?

testing/e2e/suites/osm.go
```diff
@@ -0,0 +1,64 @@
...
+						return nil
+					}); err != nil {
+					return err
+				}
+
+				return nil
+			},
+		},
+		{
+			name: "osm service",
```
jaiveerk, line 41: will we remove this when we fully shift to expected behavior where the client should create their own ingress? don't wanna have too much bloat in the e2e suite

jaiveerk (changes requested):

.github/workflows/e2ev2-provision-test.yaml
OliverMKing: failure, success, skipped. The intent is to only run this if the previous command ran (because that's what created the log we are `cat`ing). 

I will note that the `always()` function is special too. It's not like `true` like you might think (that would be redundant to have a `true` && (other expression)). It makes it so this step will run even if the job's overall status is cancelled (which would happen if the previous step failed).

testing/e2e/manifests/common.go
OliverMKing: `go mod tidy` installs dependencies as well. We check for if the thing we are trying to run has any external dependencies that need to be installed and if so we add that "install" command to the script. Go mod tidy just happens to be how we are choosing to install in this case.

testing/e2e/suites/osm.go
OliverMKing: sure. We have to have tests for it until then.

testing/e2e/infra/infras.go
```diff
@@ -10,6 +12,11 @@ var (
 	location = "South Central US"
 )
 
+var (
+	// OsmInfraName is the name of the infrastructure that will be used to test an OSM enable cluster
```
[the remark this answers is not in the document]
OliverMKing: refactored

testing/e2e/manifests/embedded/prom.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		for _, target := range targets.Active {
+			if target.Labels[jobKey] == nginxJob {
+				log.Print("found active nginx-ingress target")
+				return
+			}
+		}
+
+		log.Print("no active nginx-ingress targets found")
```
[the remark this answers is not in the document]
OliverMKing: switched it up to check the status as well which does check that metrics are actively flowing

testing/e2e/suites/operatorConfig.go
```diff
@@ -24,14 +24,25 @@ type cfgBuilderWithOsm struct {
 	osmEnabled []bool
 }
 
-func (c cfgBuilder) withOsm(enabled ...bool) cfgBuilderWithOsm {
+func (c cfgBuilder) withOsm(in infra.Provisioned, enabled ...bool) cfgBuilderWithOsm {
```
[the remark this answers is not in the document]
OliverMKing: refactored to give a single source of truth for mc opts.

testing/e2e/manifests/common.go
bfoley13: Exactly, so you can use a single command definition that includes `go mod tidy` and remove the package checks.  There wouldnt be any real performance hits if something doesnt have external dependencies

.github/workflows/e2ev2-provision-test.yaml
bfoley13: then id nit this to be swapped to `steps.test.outcome != 'skipped'`

OliverMKing: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

there's also cancelled.

OliverMKing: Forgot about that one. I'd rather be explicit about what we accept. Doing the inverse would make this incorrect if they add new statuses in the future.

bfoley13 (approved):

jaiveerk (approved):

PR #109, opened 2023-10-05
Davidgamero/add service principal infrastructure

davidgamero:
# Description

Add support for a new `infra` type with service principal support for e2e testing

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ X] Existing infra deployments succeed
- [ X] New infra deployment type succeeds

# Checklist:


Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: clean dev push push-tester-image e2e run-e2e
 
--include .env
+include .env
```
jaiveerk, line 3: nit can you merge master

Makefile
```diff
@@ -23,9 +23,15 @@ push:
...
 	# parenthesis preserve current working directory
 	(cd testing/e2e && \
-	 go run ./main.go infra --subscription=${SUBSCRIPTION_ID} --tenant=${TENANT_ID} --names=${INFRA_NAMES} && \
+	 go run ./main.go infra --subscription=${SUBSCRIPTION_ID} --tenant=${TENANT_ID} --names=${INFRA_NAMES} --sp-app-obj-id=${SERVICE_PRINCIPAL_APP_OBJ_ID} && \
 	 go run ./main.go deploy)
 
+
+e2e-infra-only:
+	# parenthesis preserve current working directory
+	(cd testing/e2e && \
```
jaiveerk, line 32: this just does what make dev does right?

testing/e2e/clients/aks.go
```diff
@@ -76,6 +91,37 @@ var OsmClusterOpt = McOpt{
 	},
 }
 
+var ServicePrincipalClusterOpt = McOpt{
+	Name: "service principal cluster",
+	fn: func(mc *armcontainerservice.ManagedCluster, opt McOptFields) error {
```
jaiveerk, line 96: I thought you were gonna take the approach of having a mcOpt generator that would use the external argument and return a function that used those? so in this case spOpts would just be passed into the mcOpt generator?

jaiveerk (changes requested):

.env.example
```diff
@@ -2,4 +2,5 @@
 
 TENANT_ID=<azure_tenant_id>
 SUBSCRIPTION_ID=<azure_subscription id>
-INFRA_NAMES="basic cluster"
\ No newline at end of file
+INFRA_NAMES="basic cluster"
+SERVICE_PRINCIPAL_APP_OBJ_ID="<azure_app_registration_object_id>"
```
OliverMKing, line 6: we should have some details on how to generate this in /docs

OliverMKing: or maybe even a script that generates it for us

Makefile
OliverMKing: this isn't a replacement for what make dev does. we have no tooling for using the infra created by this manually yet. 

we should remove this make target. keep make clean with straightforward targets that are useful.

This is basically mostly for if you want to test e2e infra changes you made yourself in which case you could just run `(cd testing/e2e && \
	 go run ./main.go infra --subscription=${SUBSCRIPTION_ID} --tenant=${TENANT_ID} --names=${INFRA_NAMES} --sp-app-obj-id=${SERVICE_PRINCIPAL_APP_OBJ_ID})`

.env.example
davidgamero: good point i'll add some

Makefile
davidgamero: yeah

Makefile
davidgamero: yeah i just used that for testing i'll take it out

testing/e2e/clients/aks.go
davidgamero: that was the plan, but we need access to the service principal fields in the `provision` function, which calls the McOpt.fn function, so i just passed it there to remove the extra layer in the call stack. Do you think adding a constructor for it would be cleaner?
i ended up passing the appObjectId as a flag instead of an environment var which made me question the constructor approach

.env.example
davidgamero: also we only need one of these per tenant, which will realistically map to one per cloud

testing/e2e/infra/types.go
```diff
@@ -81,6 +83,7 @@ type cert interface {
 type Provisioned struct {
 	Name              string
 	Cluster           cluster
+	SPCluster         cluster
```
OliverMKing, line 86: Cluster should have a consistent interface. We shouldn't add SPCluster here. Make the SPCluster fit under the Cluster inteface

testing/e2e/infra/provision.go
```diff
@@ -124,6 +137,11 @@ func (i *infra) Provision(ctx context.Context, tenantId, subscriptionId string)
 	// connect permissions
 	var permEg errgroup.Group
 
+	roleAssignmentPrincipalId := ret.Cluster.GetPrincipalId()
+	if i.ServicePrincipalOptions != nil {
```
OliverMKing, line 141: why do we need this logic here instead of building it into the aks client? GetPrincipalId() can return the necessary principalId itself. We should move this logic there

testing/e2e/infra/types.go
davidgamero: done

testing/e2e/clients/aks.go
```diff
@@ -38,16 +40,29 @@ type aks struct {
...
+type ServicePrincipalOptions struct {
+	ApplicationObjectID          string
+	ApplicationClientID          string
+	ServicePrincipalObjectID     string
+	ServicePrincipalCredPassword string
+}
+
+type McOptFields struct {
+	ClusterName             string
+	Ctx                     context.Context
```
OliverMKing, line 52: we shouldn't store ctx in a struct. Let's move it to the function definition level.

> // Programs that use Contexts should follow these rules to keep interfaces
// consistent across packages and enable static analysis tools to check context
// propagation:
//
// Do not store Contexts inside a struct type; instead, pass a Context
// explicitly to each function that needs it. The Context should be the first
// parameter, typically named ctx:
//
//	func DoSomething(ctx context.Context, arg Arg) error {
//		// ... use ctx ...
//	}

testing/e2e/clients/aks.go
```diff
@@ -38,16 +40,29 @@ type aks struct {
...
+	ServicePrincipalOptions *ServicePrincipalOptions
```
OliverMKing, line 53: I'm not 100% sure sp should be a mcoption. Maybe it should just be another parameter of the NewAKS function. 

`func NewAks(ctx context.Context, subscriptionId, resourceGroup, name, location string, spOpts *ServicePrincipalOptions, mcOpts ...McOpt) (*aks, error) {`

This function definition is pretty rough. We define spOpts in the function directly and in the mcOpts? That's not clean. Choose one place to define and do it well.

testing/e2e/clients/application.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/Azure/aks-app-routing-operator/pkg/util"
+	"github.com/Azure/aks-app-routing-operator/testing/e2e/logger"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	msgraphsdk "github.com/microsoftgraph/msgraph-sdk-go"
+	graphapplications "github.com/microsoftgraph/msgraph-sdk-go/applications"
+	graphmodels "github.com/microsoftgraph/msgraph-sdk-go/models"
+)
+
+// GetServicePrincipalOptions populates a new ServicePrincipalOptions struct with fresh credentials and application/client/servicePrincipal object ids
+func GetServicePrincipalOptions(ctx context.Context, applicationObjectID string, credName string) (*ServicePrincipalOptions, error) {
```
OliverMKing, line 17: What is ServicePrincipalOptions?

Why is this not just called get ServicePrincipal?

testing/e2e/clients/aks.go
davidgamero: agreed i just removed it we don't need it in any McOpts anymore, we can reintroduce it if needed

testing/e2e/clients/aks.go
davidgamero: agreed. moved it out from mcoptions completely

testing/e2e/clients/application.go
davidgamero: options is not a descriptive name. it has a new set of creds, and information about the containing application which isn't part of the service principal either so it has more than just a service principal

davidgamero: i guess all of those are relevant to the sp, so we could call it that anyway

testing/e2e/clients/aks.go
```diff
@@ -38,6 +39,13 @@ type aks struct {
 	options                             map[string]struct{}
 }
 
+type ServicePrincipal struct {
+	ApplicationObjectID          string
```
davidgamero, line 43: readonly for fields we dont add

testing/e2e/infra/provision.go
```diff
@@ -67,7 +80,7 @@ func (i *infra) Provision(ctx context.Context, tenantId, subscriptionId string)
 	})
 
 	resEg.Go(func() error {
-		ret.Cluster, err = clients.NewAks(ctx, subscriptionId, i.ResourceGroup, "cluster"+i.Suffix, i.Location, i.McOpts...)
+		ret.Cluster, err = clients.NewAks(ctx, subscriptionId, i.ResourceGroup, "cluster"+i.Suffix, i.Location, i.ServicePrincipalOptions, i.McOpts...)
```
davidgamero, line 83: jkariya: rename to ServicePrincipal

testing/e2e/infra/types.go
```diff
@@ -22,6 +23,7 @@ type infra struct {
 	// for resources to be provisioned inside
 	ResourceGroup, Location string
 	McOpts                  []clients.McOpt
+	ServicePrincipalOptions *clients.ServicePrincipal
```
davidgamero, line 26: jkatariya: rename

testing/e2e/clients/aks.go
```diff
@@ -157,8 +173,9 @@ func NewAks(ctx context.Context, subscriptionId, resourceGroup, name, location s
 	if result.ManagedCluster.Properties == nil {
 		return nil, fmt.Errorf("managed cluster properties is nil")
 	}
-	if result.ManagedCluster.Properties.IdentityProfile == nil {
-		return nil, fmt.Errorf("managed cluster identity profile is nil")
+	// cluster must use either MSI or Service Principal
+	if result.ManagedCluster.Properties.IdentityProfile == nil && result.ManagedCluster.Properties.ServicePrincipalProfile == nil {
```
davidgamero, line 177: add check for both

testing/e2e/clients/aks.go
```diff
@@ -133,6 +141,14 @@ func NewAks(ctx context.Context, subscriptionId, resourceGroup, name, location s
 		},
 	}
 
+	// apply service principal
+	if spOpts != nil {
+		mc.Properties.ServicePrincipalProfile = &armcontainerservice.ManagedClusterServicePrincipalProfile{
```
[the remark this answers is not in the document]
davidgamero: easier to just add Identity when using it in MSI, no need to nil out anything

testing/e2e/clients/aks.go
```diff
@@ -167,29 +184,45 @@ func NewAks(ctx context.Context, subscriptionId, resourceGroup, name, location s
...
+		if identity.ObjectID == nil {
+			return nil, fmt.Errorf("kubelet identity object id is nil")
+		}
+		if identity.ClientID == nil {
+			return nil, fmt.Errorf("kubelet identity client id is nil")
+		}
+		principalID = *identity.ObjectID
+		clientID = *identity.ClientID
+	} else {
+		principalID = spOpts.ServicePrincipalObjectID
```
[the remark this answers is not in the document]
davidgamero: a: no

.env.example
davidgamero: added

testing/e2e/clients/aks.go
```diff
@@ -90,7 +102,7 @@ func LoadAks(id azure.Resource, dnsServiceIp, location, principalId, clientId st
 	}
 }
 
-func NewAks(ctx context.Context, subscriptionId, resourceGroup, name, location string, mcOpts ...McOpt) (*aks, error) {
+func NewAks(ctx context.Context, subscriptionId, resourceGroup, name, location string, spOpts *ServicePrincipal, mcOpts ...McOpt) (*aks, error) {
```
OliverMKing, line 105: Add a comment to this explaining that if spOpts == nil we use managed identity

testing/e2e/clients/application.go
```diff
@@ -0,0 +1,58 @@
...
+	getAppResponse, err := graphClient.Applications().ByApplicationId(applicationObjectID).Get(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("getting application with object id %s: %w", applicationObjectID, err)
+	}
+
+	// add new password credential
+	addPasswordReq := graphapplications.NewItemAddPasswordPostRequestBody()
+	newCreds := graphmodels.NewPasswordCredential()
+	newCreds.SetDisplayName(util.StringPtr(credName))
+	newCreds.SetEndDateTime(to.Ptr(time.Now().Add(24 * time.Hour)))
```
OliverMKing, line 37: we should limit this to the same amount of time we keep an rg to? 2 hours

testing/e2e/clients/application.go
```diff
@@ -0,0 +1,58 @@
...
+	addPasswordReq.SetPasswordCredential(newCreds)
+	addPasswordCredResp, err := graphClient.Applications().ByApplicationId(applicationObjectID).AddPassword().Post(ctx, addPasswordReq, nil)
```
OliverMKing, line 39: what permissions does the application have? We should limit it as much as possible. Maybe document the permissions it should have in doc page?

testing/e2e/infra/provision.go
```diff
@@ -30,6 +31,18 @@ func (i *infra) Provision(ctx context.Context, tenantId, subscriptionId string)
...
+	if i.ServicePrincipal != nil {
+		if applicationObjectId == "" {
+			return ret, logger.Error(lgr, fmt.Errorf("application object id must be provided when provisioning infrastructure with service principal options"))
+		}
+		credName := i.Name + "-cred"
+		spOpt, err := clients.GetServicePrincipalOptions(ctx, applicationObjectId, credName)
+		if err != nil {
+			return ret, logger.Error(lgr, fmt.Errorf("getting app with credential: %w", err))
+		}
+		i.ServicePrincipal = spOpt
```
OliverMKing, line 43: it's a weird pattern that we check if `i.ServicePrincipal != nil` when entering this block then overwrite it. Maybe the field on infra should be called useSp or something?

davidgamero: good catch, went with an enum to make it clearer

testing/e2e/clients/application.go
davidgamero: the application itself doesn't set perms on creation, so that's down to the service principal later

testing/e2e/clients/aks.go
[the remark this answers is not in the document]
jaiveerk: nonetheless we'll want to make sure we act accordingly in the acr context

testing/e2e/clients/aks.go
jaiveerk: can u add the comments in a per field basis so that they show up in IDEs when accessed/when ppl hover over them

jaiveerk (changes requested):

jaiveerk (approved):

OliverMKing (approved):

PR #122, opened 2023-10-31
apply crds on operator startup

OliverMKing:
# Description

Applies the crds on operator startup, crashes the operator if it can't do that.

## Type of change

# How Has This Been Tested?

Unit and manually. Will be e2e tested with future additions since tests will rely on crds which will rely on the definition being in the cluster.

# Checklist:

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -7,11 +7,13 @@ import (
 	"context"
 	"net/http"
 
+	approutingv1alpha1 "github.com/Azure/aks-app-routing-operator/api/v1alpha1"
 	"github.com/go-logr/logr"
 	cfgv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
 	ubzap "go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


pkg/controller/controller_test.go
```diff
@@ -96,9 +75,3 @@ func TestGetSelfDeploy(t *testing.T) {
 		require.Nil(t, self)
 	})
 }
-
-func TestNewManagerForRestConfig(t *testing.T) {
```
OliverMKing, line 100: can't test things that use the utils.upsert function with the fake environment. That's fine. NewManagerForRestConfig is essentially the main function and doesn't have any logic by itself. Isn't valuable to unit test in the first place. 

This will be covered by all e2e tests as well.

pkg/controller/controller.go
```diff
@@ -84,17 +91,26 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
...
+	setupLog := m.GetLogger().WithName("setup")
+	deploy, err := getSelfDeploy(kcs, conf, setupLog)
 	if err != nil {
 		return nil, err
 	}
-	log.V(2).Info("using namespace: " + conf.NS)
+	setupLog.V(2).Info("using namespace: " + conf.NS)
+
+	if err := loadCRDs(cl, conf, setupLog); err != nil {
+		return nil, err
```
jaiveerk, line 112: can we log this error here or at least in NewManager

jaiveerk (changes requested):
minor change but lgtm

pkg/controller/controller.go
OliverMKing: https://github.com/Azure/aks-app-routing-operator/blob/d6fd1749935482477a7b2f88ab76dafdac750b6f/main.go#L24

OliverMKing: same way all errors in NewManager are shown

jaiveerk (approved):

gambtho (approved):

PR #130, opened 2023-11-08
add new events on published to Ingress in case of reconcile failures

OliverMKing:
# Description

Add some new events that are published to cx Ingresses utilizing App Routing when we fail to upsert resources needed for the Ingress to function. Provides significantly increased customer visibility. 

We publish to the Ingress object instead of the NginxIngressController crd because
- Ingress is the source of why these failures occur
- NIC crd is out of scope for these errors. Something like the IngressBackend failing to upsert isn't necessarily the result of the NIC, it could be bad cx configuration on the Ingress. We don't want to spam NIC with events that aren't related to it. Makes sense to send events to the customer managed resource that is closest to the reason why something isn't working.

## Type of change

# How Has This Been Tested?

Tested locally

# Checklist:

### DIFF
--- a/pkg/controller/keyvault/ingress_secret_provider_class.go
+++ b/pkg/controller/keyvault/ingress_secret_provider_class.go
@@ -112,6 +112,9 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
 	if ok {
 		logger.Info("reconciling secret provider class for ingress")
 		err = util.Upsert(ctx, i.client, spc)
+		if err != nil {
+			i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
+		}
 		return result, err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -112,6 +112,9 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
 	if ok {
 		logger.Info("reconciling secret provider class for ingress")
 		err = util.Upsert(ctx, i.client, spc)
+		if err != nil {
```
OliverMKing, line 115: these are hard to unit test because it's hard to simulate what causes upsert errors (primarily admission webhooks / policy webhooks) with the fake client.

I think it's okay not to unit test for now because this is a pretty straightforward way of using native k8s code. we are just using the k8s packages to do their intended purposes

davidgamero (approved):
lgtm

gambtho (approved):

bfoley13 (approved):

PR #155, opened 2024-01-23
adds tls reconciler

OliverMKing:
# Description

Adds a tls reconciler that adds the TLS portion to Ingresses that have the keyvault tls cert managed by us. We reconcile the Ingress object to point to the managed secret that contains the cert pulled from keyvault if the Ingress contains a specific annotation.

To use this feature, users will need to add the `kubernetes.azure.com/tls-cert-keyvault-managed: true` annotation to their Ingress.

Essentially, the reconcile logic is as follows:
- Do nothing if the Ingress' IngressClass is not an App Routing IngressClass
- Do nothing if the Ingress doesn't contain the `kubernetes.azure.com/tls-cert-keyvault-managed: true` annotation
- If the Ingress has the `kubernetes.azure.com/tls-cert-keyvault-managed: true` annotation but doesn't contain the `kubernetes.azure.com/tls-cert-keyvault-uri: <uri>` annotation then we push an Event to the Ingress informing them that we can't manage TLS without that.
- If the Ingress contains both required annotations we always set the TLS section of the Ingress to point to the App Routing managed Secret. We also update the TLS hosts to exactly match the hosts section of the rules part of the Ingress spec. We do this regardless of if the TLS section was empty before. This is a truly "managed" experience. If we chose not to reconcile if the TLS section was non-empty then we wouldn't be able to handle the case of a user adding a new Rule and hostname to the spec of the Ingress.

## Type of change

# How Has This Been Tested?

unit

# Checklist:


pkg/controller/keyvault/ingress_secret_provider_class_test.go
```diff
@@ -469,4 +469,9 @@ func TestIngressSecretProviderClassReconcilerBuildSPCFailedIsManaging(t *testing
 	require.False(t, ok)
 	require.Error(t, err)
 	require.ErrorContains(t, err, "determining if ingress is managed")
-}
\ No newline at end of file
+}
+
+func TestCertSecretName(t *testing.T) {
```
jaiveerk, line 474: love the test coverage

jaiveerk (approved):

sabbour (dismissed):
Approving the logic. Looks good, thank you.

aamgayle (approved):
LGTM

pkg/controller/keyvault/ingress_tls.go
```diff
@@ -0,0 +1,104 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	if _, ok := ing.Annotations[tlsCertKvUriAnnotation]; !ok {
+		logger.Info("ingress does not have keyvault annotation")
+		i.events.Eventf(ing, "Warning", "KeyvaultUriAnnotationMissing", "Ingress has %[1]s annotation but is missing %[2]s annotation. %[2]s annotation is needed to manage Ingress TLS.", tlsCertManagedAnnotation, tlsCertKvUriAnnotation)
+		return ctrl.Result{}, nil
+	}
+
+	logger.Info("adding TLS spec to ingress")
```
bfoley13, line 86: What if a TLS definition already exists? I think we at least need some logging in here that we're overwriting existing TLS defintions

OliverMKing: seems fair to log it but we do need to override them.

> If the Ingress contains both required Ingresses we always set the TLS section of the Ingress to point to the App Routing managed Secret. We also update the TLS hosts to exactly match the hosts section of the rules part of the Ingress spec. We do this regardless of if the TLS section was empty before. This is a truly "managed" experience. If we chose not to reconcile if the TLS section was non-empty then we wouldn't be able to handle the case of a user adding a new Rule and hostname to the spec of the Ingress.

otherwise, we wouldn't be able to handle them just adding new hosts to the rules

bfoley13: And what about rules with no hosts? Is it possible to create an ingress with a rule that has no host but has the TLS host be provided? The host replacement below would break on this case

OliverMKing: k8s docs say https://kubernetes.io/docs/concepts/services-networking/ingress/#tls

> hosts in the tls section need to explicitly match the host in the rules section.

So, it's not possible for an Ingress to have a no host rule but have the tls host. 

Your point is good though. I'll change the logic to ensure that host is actually defined for the rule because it's possible for it not to be for some routes.

OliverMKing: added log + handle rule with no host

davidgamero:
in the description,
> If the Ingress contains both required Ingresses
this is referring to annotations right?

OliverMKing:
> in the description,
> 
> > If the Ingress contains both required Ingresses
> 
> this is referring to annotations right?

yes

davidgamero (approved):
lgtm

jaiveerk (approved):

PR #166, opened 2024-02-15
defaultSSLCertificate via key vault URI

aamgayle:
# Description
This will allow users to set a default SSL Certificate via keyvault URI. It uses a new reconciler that watches the CRD for the addition of a keyvault URI under defaultSSLCertificate and creates and points the nginx manifest to an associated SPC afterward. 

Fixes # (issue)
Feature # (details)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] Unit Tests
- [X] e2e Tests
- [X] testing in an actual cluster

# Checklist:


api/v1alpha1/nginxingresscontroller_types.go
```diff
@@ -61,6 +61,10 @@ type DefaultSSLCertificate struct {
 	// Secret is a struct that holds the name and namespace fields used for the default ssl secret
 	// +optional
 	Secret *Secret `json:"secret,omitempty"`
+
+	// Secret in the form of a Key Vault URI
+	// +optional
+	KeyVaultURI string `json:"keyVaultURI"`
```
OliverMKing, line 67: this is optional so should be represented by a pointer.

api/v1alpha1/nginxingresscontroller_types.go
```diff
@@ -61,6 +61,10 @@ type DefaultSSLCertificate struct {
 	// Secret is a struct that holds the name and namespace fields used for the default ssl secret
```
OliverMKing, line 61: we need cel validation guaranteeing that only one of secret and keyvaulturi are filled out

pkg/controller/keyvault/nginx_secret_provider_class.go
```diff
@@ -0,0 +1,223 @@
...
+	logger = logger.WithValues("name", nic.Name, "namespace", "app-routing-system", "generation", nic.Generation)
+
+	spc := &secv1.SecretProviderClass{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "secrets-store.csi.x-k8s.io/v1",
+			Kind:       "SecretProviderClass",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DefaultNginxCertName(nic),
+			Namespace: "app-routing-system",
```
OliverMKing, line 93: we should be taking the namespace through the config. https://github.com/Azure/aks-app-routing-operator/blob/main/pkg/config/config.go

example of another controller doing this https://github.com/Azure/aks-app-routing-operator/blob/aa4f99d481269d43a42df4f9e8ee785b60ab394c/pkg/controller/nginxingress/nginx_ingress_controller.go#L51

pkg/controller/keyvault/nginx_secret_provider_class.go
```diff
@@ -0,0 +1,223 @@
...
+	if manifests.HasTopLevelLabels(toCleanSPC.Labels) {
+		logger.Info("removing secret provider class for ingress")
+		err = i.client.Delete(ctx, toCleanSPC)
+		return result, client.IgnoreNotFound(err)
+	}
+
+	return result, nil
+}
+
+func (i *NginxSecretProviderClassReconciler) buildSPC(nic *approutingv1alpha1.NginxIngressController, spc *secv1.SecretProviderClass) (bool, error) {
```
OliverMKing, line 139: this should likely be a more generic function. 

it's almost exactly the same as here

https://github.com/Azure/aks-app-routing-operator/blob/aa4f99d481269d43a42df4f9e8ee785b60ab394c/pkg/controller/keyvault/ingress_secret_provider_class.go#L140

we should be able to merge them together and use the same fn for both

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,49 +110,92 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+				return result, client.IgnoreNotFound(err)
+			}
 		}
-	}
 
-	managed, err := p.ingressManager.IsManaging(ing)
-	if err != nil {
-		return result, fmt.Errorf("determining if ingress is managed: %w", err)
-	}
+		if nic.Name == "" || nic.Spec.IngressClassName == "" || nic.Spec.DefaultSSLCertificate.KeyVaultURI == "" {
```
OliverMKing, line 124: why nic.Spec.IngressClassName == ""?

OliverMKing: i don't think we need to check the nic.Spec.DefaultSSLCertificate.KeyVaultURI either

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,49 +110,92 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+			logger.Info("cleaning unused placeholder pod deployment")
+			logger.Info("getting placeholder deployment")
+			toCleanDeployment := &appsv1.Deployment{}
```
OliverMKing, line 127: i'm not sure about this cleanup logic. we should be cleaning up if the spc is removed. the spc is removed when the NginxIngressController resource no longer needs it.

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,49 +110,92 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+		logger.Info("reconciling placeholder deployment for secret provider class")
+		if err = p.buildDeployment(ctx, dep, spc, nic.Name); err != nil {
+			err = fmt.Errorf("building deployment: %w", err)
+			p.events.Eventf(nic, "Warning", "FailedUpdateOrCreatePlaceholderPodDeployment", "error while building placeholder pod Deployment needed to pull Keyvault reference: %s", err.Error())
+			logger.Error(err, "failed to build placeholder deployment")
+			return result, err
+		}
+
+		if err = util.Upsert(ctx, p.client, dep); err != nil {
+			p.events.Eventf(nic, "Warning", "FailedUpdateOrCreatePlaceholderPodDeployment", "error while creating or updating placeholder pod Deployment needed to pull Keyvault reference: %s", err.Error())
```
OliverMKing, line 148: we should log the error

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,49 +110,92 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return result, err
+		}
+	} else {
```
OliverMKing, line 151: pretty much all this branching logic is exactly the same except one is for pushing events to an ingress resource and the other is to nic resources. Instead of having an if else here we should have a function that gets the responsible object then just performs the exact same logic for both. 

nic and ingress are both client.Object interfaces so you can treat them the same

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -189,7 +234,7 @@ func (p *PlaceholderPodController) buildDeployment(ctx context.Context, dep *app
 				Annotations: map[string]string{
 					"kubernetes.azure.com/observed-generation": strconv.FormatInt(spc.Generation, 10),
 					"kubernetes.azure.com/purpose":             "hold CSI mount to enable keyvault-to-k8s secret mirroring",
-					"kubernetes.azure.com/ingress-owner":       ing.Name,
+					"kubernetes.azure.com/ingress-owner":       ingName,
```
OliverMKing, line 237: using the annotation ingress-owner for the nic doesn't make sense to me

pkg/controller/nginxingress/nginx_ingress_controller.go
```diff
@@ -527,9 +528,13 @@ func ToNginxIngressConfig(nic *approutingv1alpha1.NginxIngressController, defaul
...
 
-	if nic.Spec.DefaultSSLCertificate != nil &&
-		nic.Spec.DefaultSSLCertificate.Secret.Name != "" && nic.Spec.DefaultSSLCertificate.Secret.Namespace != "" {
-		nginxIng.DefaultSSLCertificate = nic.Spec.DefaultSSLCertificate.Secret.Namespace + "/" + nic.Spec.DefaultSSLCertificate.Secret.Name
+	if nic.Spec.DefaultSSLCertificate != nil {
+		if nic.Spec.DefaultSSLCertificate.Secret != nil && nic.Spec.DefaultSSLCertificate.Secret.Name != "" && nic.Spec.DefaultSSLCertificate.Secret.Namespace != "" {
+			nginxIng.DefaultSSLCertificate = nic.Spec.DefaultSSLCertificate.Secret.Namespace + "/" + nic.Spec.DefaultSSLCertificate.Secret.Name
+		}
+		if nic.Spec.DefaultSSLCertificate != nil && nic.Spec.DefaultSSLCertificate.KeyVaultURI != "" {
+			nginxIng.DefaultSSLCertificate = "app-routing-system/" + keyvault.DefaultNginxCertName(nic)
```
OliverMKing, line 536: app-routing-system namespace shouldn't be hardcoded. consume from the config

testing/e2e/suites/nginxIngressController.go
```diff
@@ -112,6 +112,15 @@ func nicTests(in infra.Provisioned) []test {
...
 
+				testNIC = manifests.NewNginxIngressController("nginx-ingress-controller", "nginxingressclass")
+				testNIC.Spec.DefaultSSLCertificate = &v1alpha1.DefaultSSLCertificate{
+					KeyVaultURI: "Invalid.URI",
+				}
+				lgr.Info("creating NginxIngressController with invalid Secret field")
+				if err := c.Create(ctx, testNIC); err == nil {
+					return fmt.Errorf("able to create NginxIngressController despite invalid key vault uri'%s'", testNIC.Spec.ControllerNamePrefix)
+				}
+
```
OliverMKing, line 123: this e2e test only checks with an invalid keyvault uri. we need tests checking the positive scenario

api/v1alpha1/nginxingresscontroller_types.go
OliverMKing, line 67: we should validate the keyvault uri with cel

OliverMKing (changes requested):

testing/e2e/suites/nginxIngressController.go
```diff
@@ -112,6 +112,26 @@ func nicTests(in infra.Provisioned) []test {
...
+				invalidUri := "Invalid.URI"
+				testNIC = manifests.NewNginxIngressController("nginx-ingress-controller", "nginxingressclass")
+				testNIC.Spec.DefaultSSLCertificate = &v1alpha1.DefaultSSLCertificate{
+					KeyVaultURI: &invalidUri,
+				}
+				lgr.Info("creating NginxIngressController with invalid keyvault uri field")
+				if err := c.Create(ctx, testNIC); err == nil {
+					return fmt.Errorf("able to create NginxIngressController despite invalid key vault uri'%s'", testNIC.Spec.ControllerNamePrefix)
+				}
```
OliverMKing, line 133: we need an e2e test for the positive test case proving it's able to pull from kv

testing/e2e/suites/nginxIngressController.go
```diff
@@ -112,6 +112,26 @@ func nicTests(in infra.Provisioned) []test {
...
+				validUri := "https://testvault.vault.azure.net/certificates/testcert/f8982febc6894c0697b884f946fb1a34"
+				testNIC = manifests.NewNginxIngressController("nginx-ingress-controller", "nginxingressclass")
+				testNIC.Spec.DefaultSSLCertificate = &v1alpha1.DefaultSSLCertificate{
+					KeyVaultURI: &validUri,
+				}
+				lgr.Info("creating NginxIngressController with valid keyvault uri field")
+				if err := c.Create(ctx, testNIC); err != nil {
+					return fmt.Errorf("unable to create NginxIngressController despite valid key vault uri'%s'", testNIC.Spec.ControllerNamePrefix)
+				}
+
```
OliverMKing, line 124: test case for having multiple fields of the DefaultSSLCertificate filled out?

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,7 +97,19 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
 	}
 	logger = logger.WithValues("spc", spc.Name)
-	ok, err := i.buildSPC(ing, spc)
+
+	// Checking if we manage the ingress. All false cases without an error are assumed that we don't manage it
+	ok, err := i.ingressManager.IsManaging(ing)
+
+	if err != nil {
+		ok = false
+		err = fmt.Errorf("determining if ingress is managed: %w", err)
```
OliverMKing, line 106: why don't we just return early here and send the event? that's more idiomatic than waiting

pkg/controller/keyvault/ingress_secret_provider_class_test.go
```diff
@@ -337,26 +328,26 @@ func TestIngressSecretProviderClassReconcilerBuildSPCInvalidURLs(t *testing.T) {
...
-		ing := invalidURLIng.DeepCopy()
-		incorrect := "some-other-ingress-class"
-		ing.Spec.IngressClassName = &incorrect
-		ing.Annotations = map[string]string{"kubernetes.azure.com/tls-cert-keyvault-uri": "inv@lid URL"}
-
-		ok, err := i.buildSPC(ing, &secv1.SecretProviderClass{})
-		assert.False(t, ok)
-		require.NoError(t, err)
-	})
+	//t.Run("incorrect ingress class", func(t *testing.T) {
```
OliverMKing, line 336: why is this commented out?

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/Azure/aks-app-routing-operator/api/v1alpha1"
+	"github.com/Azure/aks-app-routing-operator/pkg/config"
+	kvcsi "github.com/Azure/secrets-store-csi-driver-provider-azure/pkg/provider/types"
+	v1 "k8s.io/api/networking/v1"
+	"net/url"
+	secv1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
+	"strings"
+)
+
+var NginxNamePrefix = "keyvault-nginx-"
```
OliverMKing, line 15: this should be private

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,115 @@
...
+
+func BuildSPC(ic interface{}, spc *secv1.SecretProviderClass, config *config.Config) (bool, error) {
```
OliverMKing, line 17: why is ic an interface{}?

OliverMKing: what does ic stand for? The variable name isn't clear to me

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,115 @@
...
+	var certURI, specSecretName string
+
+	if nic, ok := ic.(*v1alpha1.NginxIngressController); ok {
+		if nic.Spec.IngressClassName == "" {
```
OliverMKing, line 21: why are we checking the ingress class name here?

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,115 @@
...
+			return false, nil
+		}
+		if nic.Spec.DefaultSSLCertificate == nil || nic.Spec.DefaultSSLCertificate.KeyVaultURI == nil {
+			return false, nil
+		}
+		certURI = *nic.Spec.DefaultSSLCertificate.KeyVaultURI
+		specSecretName = DefaultNginxCertName(nic)
+	}
+
+	if ing, ok := ic.(*v1.Ingress); ok {
```
OliverMKing, line 31: the type check should probably be a switch statement with a default of returning an error due to incorrect ic type

OliverMKing: more robust and catches type errors

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,115 @@
...
+		if ing.Spec.IngressClassName == nil || ing.Annotations == nil {
```
OliverMKing, line 32: why are we checking ingressClassName here?

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,27 +108,70 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
-	if ing.Name != "" {
-		logger.Info("getting owner ingress")
-		if err = p.client.Get(ctx, client.ObjectKeyFromObject(ing), ing); err != nil {
+	return p.upsertObjectDeployment(dep, spc, req, ctx, logger)
+}
+
+func (p *PlaceholderPodController) placeholderPodCleanCheck(obj client.Object) (bool, error) {
+	nic, ok := obj.(*v1alpha1.NginxIngressController)
+	if ok {
+		if nic.Spec.DefaultSSLCertificate.KeyVaultURI == nil {
```
OliverMKing, line 117: what if nic.Spec.DefaultSSLCertificate == nil?

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,27 +108,70 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if ing.Name == "" || ing.Spec.IngressClassName == nil || !managed {
+			return true, nil
+		}
+		return false, nil
+	}
+
+	return false, nil
+}
+
+func (p *PlaceholderPodController) upsertObjectDeployment(dep *appsv1.Deployment, spc *secv1.SecretProviderClass, req ctrl.Request, ctx context.Context, logger logr.Logger) (ctrl.Result, error) {
```
OliverMKing, line 137: rename this to reconcileObjectDeployment. Upserting is inserting/updating. This handles more because it cleans

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,27 +108,70 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if objName := util.FindOwnerKind(spc.OwnerReferences, "NginxIngressController"); objName != "" {
+		obj = &v1alpha1.NginxIngressController{}
+		obj.SetName(objName)
+	}
+
+	if objName := util.FindOwnerKind(spc.OwnerReferences, "Ingress"); objName != "" {
+		obj = &netv1.Ingress{}
+		obj.SetName(objName)
+		obj.SetNamespace(req.Namespace)
+	}
```
OliverMKing, line 154: these should be a switch statement with a default of returning an error. we need to handle the case of not being able to find the owner kind

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -178,6 +223,12 @@ func (p *PlaceholderPodController) buildDeployment(ctx context.Context, dep *app
 	if old != nil { // we need to ensure that immutable fields are not changed
 		labels = old.Spec.Selector.MatchLabels
 	}
+	var ingAnnotation string
+	if _, ok := obj.(*v1alpha1.NginxIngressController); ok {
+		ingAnnotation = "kubernetes.azure.com/nginx-ingress-controller-owner"
+	} else {
+		ingAnnotation = "kubernetes.azure.com/ingress-owner"
+	}
```
OliverMKing, line 231: switch with a default case of returning error. what if obj isn't one of these two types?

testing/e2e/suites/nginxIngressController.go
OliverMKing, line 124: test the case where multiple fields in DefaultSSLCertificate are filled out

pkg/controller/keyvault/ingress_secret_provider_class_test.go
aamgayle: I meant to remove this test because it was a test case specifically for when the managed check logic was still in buildSPC. I moved that to the placeholderPodCleanCheck instead so it's not needed.

testing/e2e/suites/nginxIngressController.go
aamgayle: I didn't add one due to the CEL validation handling that but I can add that test as well!

OliverMKing: we need to test all the CEL validation.

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,7 +97,18 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
 	}
 	logger = logger.WithValues("spc", spc.Name)
-	ok, err := i.buildSPC(ing, spc)
+
+	// Checking if we manage the ingress. All false cases without an error are assumed that we don't manage it
+	ok, err := i.ingressManager.IsManaging(ing)
+
+	if err != nil {
+		return result, fmt.Errorf("determining if ingress is managed: %w", err)
+	}
```
OliverMKing, line 106: to make this more idiomatic we should have the !ok logic here (the cleanup logic).

It doesn't make sense to have multiple if ok { when we can have an early return instead

pkg/controller/keyvault/ingress_secret_provider_class_test.go
```diff
@@ -447,26 +404,12 @@ func TestIngressSecretProviderClassReconcilerBuildSPCCloud(t *testing.T) {
 	}
 }
 
-func TestIngressSecretProviderClassReconcilerBuildSPCFailedIsManaging(t *testing.T) {
-	i := &IngressSecretProviderClassReconciler{
-		ingressManager: NewIngressManagerFromFn(func(ing *netv1.Ingress) (bool, error) {
-			return false, fmt.Errorf("failed to get ingress")
-		}),
+func BuildTestSpcConfig(msiClient, tenantID, cloud string) *config.Config {
```
OliverMKing, line 407: this shouldn't be exported

pkg/controller/keyvault/ingress_secret_provider_class_test.go
```diff
@@ -316,47 +317,16 @@ func TestIngressSecretProviderClassReconcilerInvalidURL(t *testing.T) {
...
-		}),
-	}
-
 	invalidURLIng := &netv1.Ingress{
 		Spec: netv1.IngressSpec{
 			IngressClassName: &spcTestIngressClassName,
 		},
 	}
 
-	t.Run("missing ingress class", func(t *testing.T) {
```
OliverMKing, line 335: why are these test cases removed?

pkg/controller/keyvault/kv_util_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func TestDefaultNginxCertName(t *testing.T) {
+	testStr := DefaultNginxCertName(&v1alpha1.NginxIngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: maxString,
+		},
+	})
+
+	require.NotContains(t, testStr, "b")
```
OliverMKing, line 21: add a test case for more typical cases

testing/e2e/suites/utils.go
```diff
@@ -62,3 +63,11 @@ func upsert(ctx context.Context, c client.Client, obj client.Object) error {
 	lgr.Info("object updated")
 	return nil
 }
+
+func getTestSecret(name, namespace string) *v1alpha1.Secret {
```
OliverMKing, line 67: I don't understand the point of this helper function. It doesn't make anything simpler or abstract anything out

testing/e2e/suites/nginxIngressController.go
```diff
@@ -187,6 +204,101 @@ func nicTests(in infra.Provisioned) []test {
...
+				// create nic and upsert
+				testNIC := manifests.NewNginxIngressController("nginx-ingress-controller", "nginxingressclass")
+				testNIC.Spec.DefaultSSLCertificate = &defaultSSLCert
+				if err := upsert(ctx, c, testNIC); err != nil {
+					return fmt.Errorf("upserting NIC: %w", err)
+				}
+
+				var service v1alpha1.ManagedObjectReference
```
OliverMKing, line 244: why are we waiting for the service?

testing/e2e/suites/nginxIngressController.go
```diff
@@ -187,6 +204,101 @@ func nicTests(in infra.Provisioned) []test {
...
+					}
+					cleanSPC := &secv1.SecretProviderClass{}
+
+					if err := c.Get(ctx, client.ObjectKeyFromObject(spc), cleanSPC); err != nil {
+						return false, fmt.Errorf("getting SPC: %w", err)
+					} else {
+						return true, nil
+					}
+
+					lgr.Info("spc not found")
```
OliverMKing, line 289: it's impossible to ever reach this block of code. you have an if else above that always returns. we should remove the else.

testing/e2e/suites/nginxIngressController.go
OliverMKing: instead wait for the Nic to have a status of available. That should mean the spc is available too so checking the spc wouldn't have to go into a retry if you wait for the nic to be available

pkg/controller/keyvault/ingress_secret_provider_class_test.go
aamgayle: The logic to check the ingress class name was for the isManaging check. Since I moved that logic to the placeholderPodCleanCheck I moved these tests there

OliverMKing:
Looking good! Just a few small nits

api/v1alpha1/nginxingresscontroller_types.go
```diff
@@ -57,10 +57,17 @@ type NginxIngressControllerSpec struct {
 	DefaultSSLCertificate *DefaultSSLCertificate `json:"defaultSSLCertificate,omitempty"`
 }
 
+// DefaultSSLCertificate holds a secret in the form of a secret struct ith name and namespace properties or a key vault uri
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:XValidation:rule="(isURL(self.keyVaultURI) || !has(self.keyVaultURI))"
```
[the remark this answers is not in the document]
aamgayle: I do agree that it's better to have it closer to the field it deals with but I couldn't find a way to check whether or not the field itself exists or if it exists that it's a url at that level via CEL.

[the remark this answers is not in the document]
OliverMKing: makes sense. i have no problem with it as it

api/v1alpha1/nginxingresscontroller_types.go
```diff
@@ -61,10 +61,17 @@ type NginxIngressControllerSpec struct {
 	Scaling *Scaling `json:"scaling,omitempty"`
 }
 
+// DefaultSSLCertificate holds a secret in the form of a secret struct ith name and namespace properties or a key vault uri
```
OliverMKing, line 64: typo here

pkg/controller/keyvault/nginx_secret_provider_class.go
```diff
@@ -0,0 +1,135 @@
...
+			}},
+		},
+	}
+	logger = logger.WithValues("spc", spc.Name)
+	logger.Info("building spc and upserting if managed with labels")
+	upsertSPC, err := buildSPC(nic, spc, i.config)
+	if err != nil {
+		logger.Info("failed to build secret provider class for ingress, user input invalid. sending warning event")
+		i.events.Eventf(nic, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", err)
+		return result, nil
```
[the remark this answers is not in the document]
OliverMKing: as for how to fix this there's a few ways. I'm a fan of having different error types

https://stackoverflow.com/questions/38361919/how-do-i-cleanly-separate-user-facing-errors-from-internal-errors-in-golang

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +98,37 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
-	if ok {
-		logger.Info("reconciling secret provider class for ingress")
-		err = util.Upsert(ctx, i.client, spc)
-		if err != nil {
-			i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
+
+	if isManaged {
+		var upsertSPC bool
+
+		if upsertSPC, err = buildSPC(ing, spc, i.config); err != nil {
```
OliverMKing, line 111: see my other comment on the different kinds of buildSPC errors. they apply here too

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,116 @@
...
+		specSecretName = DefaultNginxCertName(t)
+	case *v1.Ingress:
+		if t.Annotations == nil {
+			return false, nil
+		}
+
+		certURI = t.Annotations["kubernetes.azure.com/tls-cert-keyvault-uri"]
+		specSecretName = fmt.Sprintf("keyvault-%s", t.Name)
+	default:
+		return false, fmt.Errorf("incorrect object type: %s", reflect.TypeOf(obj).String())
```
OliverMKing, line 38: just print t here?

avoiding reflect when possible is good. it has some weird behaviors

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,27 +108,63 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+	switch {
+	case util.FindOwnerKind(spc.OwnerReferences, "NginxIngressController") != "":
+		obj = &v1alpha1.NginxIngressController{}
+		obj.SetName(util.FindOwnerKind(spc.OwnerReferences, "NginxIngressController"))
+	case util.FindOwnerKind(spc.OwnerReferences, "Ingress") != "":
+		obj = &netv1.Ingress{}
+		obj.SetName(util.FindOwnerKind(spc.OwnerReferences, "Ingress"))
+		obj.SetNamespace(req.Namespace)
+	default:
+		return result, fmt.Errorf("failed to reconcile object deployment: object type not nginxIngressController or ingress")
```
OliverMKing, line 150: not 100% sure this is an error. we respond to lots of kinds of spc events in this reconciler, even ones not owned by us. 

If a user creates their own spc we don't want to trigger this error, it's expected and okay. probably worth logging with an info and returning nil

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +98,40 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
+				i.events.Eventf(ing, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", userErr.Error())
+				return result, nil
+			}
+			return result, err
+		}
+
+		if upsertSPC {
+			logger.Info("reconciling secret provider class for ingress")
+			if err = util.Upsert(ctx, i.client, spc); err != nil {
+				i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
```
bfoley13, line 123: should use `err.Error()`

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +98,40 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
-		if err != nil {
-			i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
+
+	if isManaged {
+		var upsertSPC bool
+
+		if upsertSPC, err = buildSPC(ing, spc, i.config); err != nil {
+			if userErr := err.(buildSPCUserError); userErr != nil {
+				logger.Info("failed to build secret provider class for ingress, user input invalid. sending warning event")
+				i.events.Eventf(ing, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", userErr.Error())
```
bfoley13, line 114: userError could be at a higher level than strictly buildSPC so in the future we should move it.  But we should include the `UserError` function which has more customer facing error responses

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +98,40 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
-		logger.Info("reconciling secret provider class for ingress")
-		err = util.Upsert(ctx, i.client, spc)
-		if err != nil {
-			i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
+
+	if isManaged {
+		var upsertSPC bool
+
+		if upsertSPC, err = buildSPC(ing, spc, i.config); err != nil {
+			if userErr := err.(buildSPCUserError); userErr != nil {
```
OliverMKing, line 112: this is the old way to check if an error is of a certain type

https://gosamples.dev/check-error-type/
use one of these instead

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,123 @@
...
+	certName := nginxNamePrefix + nic.Name
+
+	if len(certName) > secretMaxSize {
+		return certName[0:secretMaxSize]
+	}
+
+	return certName
+}
+
+type buildSPCUserError interface {
```
OliverMKing, line 117: I've never seen this pattern for a custom error before.

https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go#introduction

I prefer the struct based form

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,140 @@
...
+func certSecretName(ingressName string) string {
+	return fmt.Sprintf("keyvault-%s", ingressName)
+}
+
+type userError interface {
+	error
+	UserError() string
+}
+
+type buildSPCUserError struct {
```
bfoley13, line 126: My thought are this should contain an error as well:

`
type buildSPCUserError struct {
  err error
  userMessage string
}

// for internal use
func (b) Error() string {
  return b.err.Error()
}

// for user facing messages
func (b) UserError() string {
  return b.userMessage
}
`

bfoley13: that way the `UserError` func can be used when publishing events to the CRD

OliverMKing: I like this idea

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,149 @@
...
+func (b buildSPCUserError) UserError() string {
+	return b.userMessage
+}
+
+func newBuildSPCUserError(err error, msg string) buildSPCUserError {
+	return buildSPCUserError{err, msg}
+}
+
+var (
+	UserError         userError
```
OliverMKing, line 147: why do we need these global vars?

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,40 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
-		logger.Info("reconciling secret provider class for ingress")
-		err = util.Upsert(ctx, i.client, spc)
-		if err != nil {
-			i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
+
+	if isManaged {
+		var upsertSPC bool
+
+		if upsertSPC, err = buildSPC(ing, spc, i.config); err != nil {
+			if errors.As(err, &UserError) {
```
OliverMKing, line 113: this shouldn't be using a global var. That leads to race conditions. Just use a local var

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,40 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
+				logger.Info("failed to build secret provider class for ingress, user input invalid. sending warning event")
```
OliverMKing, line 114: even if it's a user error we should log the actual error. Include the error itself in the Info log here

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,40 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
+				i.events.Eventf(ing, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", UserError.UserError())
+				return result, nil
+			}
+			return result, err
```
OliverMKing, line 118: log the error before returning

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,41 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
 	logger = logger.WithValues("spc", spc.Name)
-	ok, err := i.buildSPC(ing, spc)
-	if err != nil {
-		logger.Info("failed to build secret provider class for ingress, user input invalid. sending warning event")
-		i.events.Eventf(ing, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", err)
-		return result, nil
+
+	// Checking if we manage the ingress. All false cases without an error are assumed that we don't manage it
+	var isManaged bool
+	if isManaged, err = i.ingressManager.IsManaging(ing); err != nil {
```
OliverMKing, line 105: need to error log this

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,41 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
+		var upsertSPC bool
+
+		if upsertSPC, err = buildSPC(ing, spc, i.config); err != nil {
+			var userErr userError
+			if errors.As(err, &userErr) {
+				logger.Info(fmt.Sprintf("failed to build secret provider class for ingress with error: %s. sending warning event"), userErr.Error())
+				i.events.Eventf(ing, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", userErr.UserError())
+				return result, nil
+			}
+			return result, err
```
OliverMKing, line 119: need to error log this

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,41 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
+		}
+
+		if upsertSPC {
+			logger.Info("reconciling secret provider class for ingress")
+			if err = util.Upsert(ctx, i.client, spc); err != nil {
+				i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err.Error())
```
OliverMKing, line 125: need to log the error

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,45 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
-	ok, err := i.buildSPC(ing, spc)
-	if err != nil {
-		logger.Info("failed to build secret provider class for ingress, user input invalid. sending warning event")
-		i.events.Eventf(ing, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", err)
-		return result, nil
+
+	// Checking if we manage the ingress. All false cases without an error are assumed that we don't manage it
+	var isManaged bool
+	if isManaged, err = i.ingressManager.IsManaging(ing); err != nil {
+		logger.Info(fmt.Sprintf("failed while checking if ingress was managed with error: %s.", err.Error()))
```
OliverMKing, line 106: this is an error level log

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,45 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
+
+			logger.Info(fmt.Sprintf("failed to build secret provider class for ingress with error: %s.", err.Error()))
+			return result, err
+		}
+
+		if upsertSPC {
+			logger.Info("reconciling secret provider class for ingress")
+			if err = util.Upsert(ctx, i.client, spc); err != nil {
+				i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err.Error())
+				logger.Info(fmt.Sprintf("failed to upsert secret provider class for ingress with error: %s.", err.Error()))
```
OliverMKing, line 129: error level log

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,144 @@
...
+func certSecretName(ingressName string) string {
+	return fmt.Sprintf("keyvault-%s", ingressName)
+}
+
+type userError interface {
+	error
+	UserError() string
+}
+
+type buildSPCUserError struct {
```
OliverMKing, line 127: this should just be a generic UserError struct but it's fine to swap to that in the future

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return certName
+}
+
+func certSecretName(ingressName string) string {
+	return fmt.Sprintf("keyvault-%s", ingressName)
+}
+
+type userError interface {
```
OliverMKing, line 122: what is this interface used for? we shouldn't use this?

pkg/controller/keyvault/ingress_secret_provider_class.go
```diff
@@ -103,28 +99,45 @@ func (i *IngressSecretProviderClassReconciler) Reconcile(ctx context.Context, re
...
-		if err != nil {
-			i.events.Eventf(ing, "Warning", "FailedUpdateOrCreateSPC", "error while creating or updating SecretProviderClass needed to pull Keyvault reference: %s", err)
+
+	if isManaged {
+		var upsertSPC bool
+
+		if upsertSPC, err = buildSPC(ing, spc, i.config); err != nil {
+			var userErr userError
+			if errors.As(err, &userErr) {
+				logger.Error(err, fmt.Sprintf("failed to build secret provider class for ingress with error: %s. sending warning event", userErr.Error()))
```
OliverMKing, line 116: this is info

pkg/controller/keyvault/kv_util.go
```diff
@@ -0,0 +1,144 @@
...
+func (b buildSPCUserError) Error() string {
+	return b.err.Error()
+}
+
+// for user facing messages
+func (b buildSPCUserError) UserError() string {
+	return b.userMessage
+}
+
+func newBuildSPCUserError(err error, msg string) buildSPCUserError {
```
OliverMKing, line 142: rename as well

pkg/controller/keyvault/nginx_secret_provider_class.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+	}
+	logger = logger.WithValues("spc", spc.Name)
+	logger.Info("building spc and upserting if managed with labels")
+	upsertSPC, err := buildSPC(nic, spc, i.config)
+
+	if err != nil {
+		var userErr userError
+		if errors.As(err, &userErr) {
+			logger.Error(err, fmt.Sprintf("failed to build secret provider class for nginx ingress controller with error: %s. sending warning event", userErr.Error()))
```
OliverMKing, line 106: info if it's user's fault

pkg/controller/keyvault/nginx_secret_provider_class.go
```diff
@@ -0,0 +1,146 @@
...
+			i.events.Eventf(nic, "Warning", "InvalidInput", "error while processing Keyvault reference: %s", userErr.UserError())
+			return result, nil
+		}
+
+		logger.Info(fmt.Sprintf("failed to build secret provider class for nginx ingress controller with error: %s.", err.Error()))
```
OliverMKing, line 111: error log

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,27 +108,64 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+	switch {
+	case util.FindOwnerKind(spc.OwnerReferences, "NginxIngressController") != "":
+		obj = &v1alpha1.NginxIngressController{}
+		obj.SetName(util.FindOwnerKind(spc.OwnerReferences, "NginxIngressController"))
+	case util.FindOwnerKind(spc.OwnerReferences, "Ingress") != "":
+		obj = &netv1.Ingress{}
+		obj.SetName(util.FindOwnerKind(spc.OwnerReferences, "Ingress"))
+		obj.SetNamespace(req.Namespace)
+	default:
+		p.events.Eventf(obj, "Waring", "FailedReconcileObjectDeployment", "failed to reconcile object deployment: object type not nginxIngressController or ingress")
```
OliverMKing, line 150: shouldn't warn. it's possible for user's to create their own spc which is completely fine and acceptable. just remove the event

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -108,27 +108,64 @@ func (p *PlaceholderPodController) Reconcile(ctx context.Context, req ctrl.Reque
...
+		return result, nil
+	}
+
+	if obj.GetName() != "" {
+		logger.Info(fmt.Sprintf("getting owner %s", obj.GetObjectKind().GroupVersionKind().Kind))
```
OliverMKing, line 155: the gvk info won't be populated here because we never set those fields ourselves and we aren't populating them with the client yet

OliverMKing: can probably just remove the kind part from here.

log which kind was found in the switch

OliverMKing (approved):
good work!

bfoley13 (approved):

PR #170, opened 2024-02-29
OSM Functionality Updates

jaiveerk:
# Description

This modifies our resource reconciliation when OSM is enabled to avoid a unique edge case encountered during RP e2e testing where client traffic failed to get through to the downstream server due to a failing OSM injector pod.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?
RP E2E, with suggestions from the OSM team.

# Checklist:


pkg/manifests/nginx.go
```diff
@@ -417,9 +417,6 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 	}
 
 	podAnnotations := map[string]string{}
-	if !conf.DisableOSM {
-		podAnnotations["openservicemesh.io/sidecar-injection"] = "enabled"
```
OliverMKing, line 421: If this is actually necessary you need to explicitly set it instead of removing it so that this is reconciled for existing resources

jaiveerk:
closing after issue was found to be on OSM's side, taking according internal action

pkg/manifests/common.go
```diff
@@ -66,6 +66,11 @@ func Namespace(conf *config.Config) *corev1.Namespace {
 		},
 	}
 
+	if !conf.DisableOSM {
+		ns.ObjectMeta.Annotations["openservicemesh.io/sidecar-injection"] = "disabled"
```
keithmattix, line 70: Are there other pods in this namespace besides NGINX that you want sidecar injected? If not, can we remove this?

jaiveerk: we don't want the sidecar injected on any of them, including nginx (right)?
but yeah I can remove the annotation. Do we also need to specify that we want the sidecar disabled on all our other pods/deployments as well?

keithmattix: The absence of the injection annotation from the workloads is sufficient for not injecting the sidecar 👍🏾

jaiveerk: should I go ahead and remove it from nginx as well? in theory, that would prevent it from restarting when this change rolls out, right?

keithmattix: IIUC, removing the annotation from NGINX pod would still trigger a reconciliation from your control plane right?

jaiveerk: ah wait nevermind - it would still restart because we're removing the annotation from each pod, so I'm fine to leave nginx as is

testing/e2e/suites/utils.go
```diff
@@ -28,7 +28,7 @@ func waitForAvailable(ctx context.Context, c client.Client, deployment appsv1.De
 			}
 		}
 
-		if time.Since(start) > 10*time.Minute { // 10 minutes because it takes a decent amount of time for dns to "propagate"
+		if time.Since(start) > 20*time.Minute { // 10 minutes because it takes a decent amount of time for dns to "propagate"
```
OliverMKing, line 31: why does this need to be bumped to 20 minutes?

testing/e2e/suites/osm.go
```diff
@@ -2,18 +2,37 @@ package suites
...
 )
 
+func deleteNginxPods(config *rest.Config) error {
+	c, err := client.New(config, client.Options{})
+	if err != nil {
+		return fmt.Errorf("creating client: %w", err)
+	}
+
+	podLabels := appManifests.GetTopLevelLabels()
+	podLabels["app.kubernetes.io/component"] = "ingress-controller"
```
OliverMKing, line 27: this shouldn't be hardcoded. export "ingress-controller"

testing/e2e/suites/osm.go
```diff
@@ -2,18 +2,37 @@ package suites
...
+	err = c.DeleteAllOf(context.Background(), &corev1.Pod{}, client.InNamespace(manifests.ManagedResourceNs), client.MatchingLabels(podLabels))
```
OliverMKing, line 28: you should pretty much always be using assignment in the if statement. that is the go pattern

testing/e2e/suites/osm.go
```diff
@@ -23,14 +42,26 @@ func osmSuite(in infra.Provisioned) []test {
...
 			run: func(ctx context.Context, config *rest.Config, operator manifests.OperatorConfig) error {
-				if err := clientServerTest(ctx, config, operator, osmNs, in,
-					func(ingress *netv1.Ingress, service *corev1.Service, _ zoner) error {
-						annotations := ingress.GetAnnotations()
-						annotations["kubernetes.azure.com/use-osm-mtls"] = "true"
-						ingress.SetAnnotations(annotations)
-
-						return nil
-					}, nil); err != nil {
+				if operator.DisableOsm {
```
OliverMKing, line 45: this isn't needed. it's specified above that osm is required

testing/e2e/suites/osm.go
```diff
@@ -44,16 +75,28 @@ func osmSuite(in infra.Provisioned) []test {
...
-					func(ingress *netv1.Ingress, service *corev1.Service, z zoner) error {
-						ingress = nil
-						annotations := service.GetAnnotations()
-						annotations["kubernetes.azure.com/ingress-host"] = z.GetNameserver()
-						annotations["kubernetes.azure.com/tls-cert-keyvault-uri"] = in.Cert.GetId()
-						service.SetAnnotations(annotations)
-
-						return nil
-					}, nil); err != nil {
+				if operator.DisableOsm {
```
OliverMKing, line 78: not needed

testing/e2e/suites/utils.go
jaiveerk: the local run I showed you succeeded, it just needed more time - believe I made this change on a call w you

testing/e2e/suites/osm.go
jaiveerk: done

testing/e2e/suites/osm.go
jaiveerk: done

testing/e2e/suites/osm.go
jaiveerk: done

testing/e2e/suites/osm.go
jaiveerk: exported

testing/e2e/suites/utils.go
OliverMKing: how much time does it need and why?

i suggested the change to see if it works but we need to figure out why. it's bad if it takes so long for things to start working after just a pod restart.

jaiveerk: Per [this form](https://learn.microsoft.com/en-us/answers/questions/524792/azure-rbac-propagation-latency) rbac propagation can take up to 30 minutes - we can add timeout as an argument to have different timeouts between when the dns contributor role is first assigned (the first testing iteration) and when nginx is restarted, but that seems like overkill

testing/e2e/clients/aks.go
```diff
@@ -357,11 +357,11 @@ func (a *aks) waitStable(ctx context.Context, objs []client.Object) error {
 					getLogsFn := func() error { // right now this just dumps all logs on the pod, if we eventually have more logs
 						// than can be stored we will need to "stream" this by using the --since-time flag
 						if err := a.runCommand(ctx, armcontainerservice.RunCommandRequest{
-							Command: to.Ptr(fmt.Sprintf("kubectl logs job/%s -n %s", obj.GetName(), ns)),
+							Command: to.Ptr(fmt.Sprintf("kubectl logs job/%s -n %s --since=5m", obj.GetName(), ns)),
```
OliverMKing, line 360: We shouldn't do this. The file is fine. The probably is on the GitHub action runner so change the workflow to tail instead of cat the entire file.

We should continue to gather the entire Job's output into this file.

OliverMKing: also we don't just want to tail. we should print the first 50 lines or so. that shows the testing order and other startup context

jaiveerk: done

.github/workflows/e2ev2-provision-test.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
 
       - name: Print test logs
         shell: bash
-        run: cat testing/e2e/job-app-routing-operator-e2e.log
+        run: head -n 100 testing/e2e/job-app-routing-operator-e2e.log && tail -n 200 testing/e2e/job-app-routing-operator-e2e.log
```
OliverMKing, line 94: print that we are printing the first 100 and last 200 logs so no one gets confused in the future that the middle is missing

jaiveerk: done

pkg/manifests/common.go
```diff
@@ -66,6 +66,10 @@ func Namespace(conf *config.Config) *corev1.Namespace {
 		},
 	}
 
+	if !conf.DisableOSM {
+		ns.ObjectMeta.Labels["openservicemesh.io/monitored-by"] = "osm"
```
OliverMKing, line 70: this mesh name is guaranteed if they are using osm addon?

jaiveerk: per the OSM team, yes. I'll double check to make sure though

testing/e2e/clients/acr.go
```diff
@@ -118,12 +120,12 @@ func (a *acr) BuildAndPush(ctx context.Context, imageName, dockerfilePath, docke
...
 			// tells us it exists then the acr command fails. This is the only reliable way we've found
 			// to wait for the acr to be ready.
-			if (!cantFindAcrRegex.Match(errLog.Bytes())) && (!throttledRegex.Match(errLog.Bytes())) {
+			if (!cantFindAcrRegex.Match(errLog.Bytes())) && (!throttledRegex.Match(errLog.Bytes())) && (!hostRegex.Match(errLog.Bytes())) {
 				lgr.Error("failed to build and push acr image")
 				return fmt.Errorf("running build and push command: %w", err)
 			}
 
-			if time.Since(start) > 1*time.Minute {
+			if time.Since(start) > 5*time.Minute {
```
OliverMKing, line 128: why is this bumped to 5 minutes? just fail faster

OliverMKing: if we have to bump this because the host regex error happens after building write custom logic for that. wait a different amount of time when getting the hostregex error vs finding regex or throttled

testing/e2e/suites/operatorConfig.go
```diff
@@ -39,11 +39,7 @@ func (c cfgBuilder) withOsm(in infra.Provisioned, enabled ...bool) cfgBuilderWit
 	for _, e := range enabled {
 		// osm tests can only work if the cluster has osm installed.
 		// filter out any enabled on clusters without osm
-		if !osmCluster && e {
-			continue
-		}
-
-		if osmCluster && !e {
+		if osmCluster != e {
```
OliverMKing, line 42: this revert isn't correct. merge with main

testing/e2e/suites/utils.go
```diff
@@ -28,7 +28,7 @@ func waitForAvailable(ctx context.Context, c client.Client, deployment appsv1.De
 			}
 		}
 
-		if time.Since(start) > 10*time.Minute { // 10 minutes because it takes a decent amount of time for dns to "propagate"
+		if time.Since(start) > 20*time.Minute { // 20 minutes because it takes a decent amount of time for dns to "propagate"
```
OliverMKing, line 31: leave the context you told me where this is no longer about dns propagating. it's about identities too

pkg/manifests/common.go
jaiveerk: confirmed with team that this won't be changing

testing/e2e/suites/utils.go
jaiveerk: done

testing/e2e/suites/operatorConfig.go
jaiveerk: reverted - but when is one ever not equal to the other? does it somehow involve a nil case?

testing/e2e/clients/acr.go
jaiveerk: reverted to failing fast

OliverMKing (approved):

davidgamero (approved):
lgtm

PR #178, opened 2024-03-15
add min and max replica configuration through crd

OliverMKing:
# Description

lets users configure min and max replicas through crd

## Type of change

# How Has This Been Tested?

Unit, e2e, and manually tested.

# Checklist:


pkg/manifests/nginx_test.go
```diff
@@ -22,6 +22,8 @@ var (
 				"service.beta.kubernetes.io/azure-load-balancer-internal": "true",
 			},
 		},
+		MinReplicas: 2,
```
jaiveerk, line 25: can you add a fixture test for when both scale values are nil? That's a possible case, right?

jaiveerk (changes requested):
lgtm just a couple of test requests

config/crd/bases/approuting.kubernetes.azure.com_nginxingresscontrollers.yaml
```diff
@@ -104,6 +104,26 @@ spec:
...
+                    type: integer
+                  minReplicas:
+                    description: MinReplicas is the lower limit for the number of
+                      Ingress Controller replicas. It defaults to 2 pods.
+                    format: int32
+                    minimum: 1
+                    type: integer
+                type: object
+                x-kubernetes-validations:
+                - rule: (!has(self.minReplicas)) || (!has(self.maxReplicas)) || (self.minReplicas
```
bosesuneha, line 125: Just a question since I'm not familiar with this: why does the rule check if minReplicas or maxReplicas are not set? Also, is there any limit for maxReplica count?

pkg/controller/nginxingress/nginx_ingress_controller_test.go
```diff
@@ -954,6 +966,145 @@ func TestToNginxIngressConfig(t *testing.T) {
...
+				ControllerClass: "approuting.kubernetes.azure.com/nicName",
+				ResourceName:    "controllerNamePrefix-0",
+				ServiceConfig:   &manifests.ServiceConfig{},
+				IcName:          "ingressClassName",
+				MaxReplicas:     47,
+				MinReplicas:     defaultMinReplicas,
+			},
+		},
+		{
+			name: "custom fields with max replica lower than min",
```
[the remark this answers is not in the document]
OliverMKing: that's covered by cel validation in the crd. we purposefully left it off here but test it in e2e

pkg/manifests/nginx_test.go
OliverMKing: that is not. the options struct used to generate manifest doesn't use a pointer.

the nil is transformed at a higher level where this is fully tested

config/crd/bases/approuting.kubernetes.azure.com_nginxingresscontrollers.yaml
OliverMKing: they are pointer values so it's possible for a user to not have them set or only one to be set. in CEL (the validation language) there's no default values so if we were to compare something null vs a defined thing it would panic.

there is no limit for max replica count. this matches the form of the hpa which this field eventually flows into

pkg/controller/nginxingress/nginx_ingress_controller.go
```diff
@@ -30,6 +30,11 @@ const (
 	controllerClassMaxLen = 250
 )
 
+const (
+	defaultMinReplicas = 2
+	defaultMaxReplicas = 100
+)
```
Vidya2606, line 36: Question: on what basis are we setting the value for min and max replicas?

OliverMKing: this is the default. It's what we previously had which is important, so the behavior doesn't change.

we have scale tested these and these are acceptable defaults

pkg/manifests/nginx_test.go
jaiveerk: got it - and that's in the controller tests right?

config/crd/bases/approuting.kubernetes.azure.com_nginxingresscontrollers.yaml
jaiveerk: if customers want scale... we'll give em scale! 😆

jaiveerk (approved):
lgtm

Vidya2606 (dismissed):

pkg/manifests/nginx_test.go
OliverMKing: exactly

config/crd/bases/approuting.kubernetes.azure.com_nginxingresscontrollers.yaml
OliverMKing: also note that this flows down to an HPA which will only scale to the max if that number is actually needed

bosesuneha: thanks, that makes sense!

PR #180, opened 2024-03-19
add scaling threshold to crd

OliverMKing:
# Description

adds scaling threshold to crd to configure how fast hpa scales. The options provided are known good values. We use enums to control the actual values which we scale test internally. We control cpu requests so that is another reason why these are enums.

## Type of change

# How Has This Been Tested?

locally, e2e, unit

# Checklist:


sabbour (dismissed):
Approved from naming/description perspective

api/v1alpha1/nginxingresscontroller_types.go
```diff
@@ -91,8 +91,24 @@ type Scaling struct {
 	// +kubebuilder:validation:Minimum=1
 	// +optional
 	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// Threshold defines how quickly the Ingress Controller pods should scale based on workload. Rapid means the Ingress Controller
+	// will scale quickly and aggressively, which is the best choice for handling sudden and significant traffic spikes. Steady
+	// is the opposite, prioritizing cost-effectiveness. Steady is the best choice when fewer replicas handling more work is desired or when
+	// traffic isn't expected to fluctuate. Balanced is a good mix between the two that works for most use-cases. It defaults to balanced.
```
jaiveerk, line 98: nit: "It defaults to balanced" is a bit unclear - can we make this Threshold defaults to balanced or "... that forks for most use-cases and is the default value"

pkg/controller/nginxingress/nginx_ingress_controller.go
```diff
@@ -560,3 +567,31 @@ func ToNginxIngressConfig(nic *approutingv1alpha1.NginxIngressController, defaul
 
 	return nginxIng
 }
+
+func getTargetCPUUtilizationPercentage(nic *approutingv1alpha1.NginxIngressController) int32 {
+	if nic == nil {
+		return defaultTargetCPUUtilization
+	}
+
+	scaling := nic.Spec.Scaling
```
jaiveerk, line 576: is it possible for nic.Spec to be nil/do we need a nil check for that as well

pkg/controller/nginxingress/nginx_ingress_controller.go
```diff
@@ -35,6 +35,12 @@ const (
 	defaultMaxReplicas = 100
 )
 
+const (
+	rapidTargetCpuUtilization   = 55
+	defaultTargetCPUUtilization = 70
```
jaiveerk, line 40: nit: can we rename to balanced? I don’t want to couple the fact that we have balanced as an option with the fact that we also have a default. Instead, we should have logic that, in the default case, returns balanced (or whatever we want it to) - not define balanced as the default to begin with

pkg/controller/nginxingress/nginx_ingress_controller_test.go
```diff
@@ -1099,12 +1111,61 @@ func TestToNginxIngressConfig(t *testing.T) {
...
+			},
+			want: manifests.NginxIngressConfig{
+				ControllerClass:                "approuting.kubernetes.azure.com/nicName",
+				ResourceName:                   "controllerNamePrefix-0",
+				ServiceConfig:                  &manifests.ServiceConfig{},
+				IcName:                         "ingressClassName",
+				MaxReplicas:                    defaultMaxReplicas,
+				MinReplicas:                    defaultMinReplicas,
+				TargetCPUUtilizationPercentage: steadyTargetCPUUtilization,
 			},
```
jaiveerk, line 1169: Can you add a custom fields with balanced threshold case as well

jaiveerk (changes requested):
few changes requested. I know we’ve got fixture tests so our bases should be covered but is there any feasible and useful way we can test for this in e2e to ensure our output values are correct all the way through to hpa deployment?

pkg/controller/nginxingress/nginx_ingress_controller.go
OliverMKing: no

api/v1alpha1/nginxingresscontroller_types.go
OliverMKing: done

pkg/controller/nginxingress/nginx_ingress_controller_test.go
OliverMKing: done

jaiveerk (approved):

bfoley13 (approved):
Changes look good

PR #186, opened 2024-03-25
add unit tests to util package

OliverMKing:
# Description

adds unit tests to util package

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -1,7 +1,9 @@
 package util
 
 import (
+	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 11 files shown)


pkg/util/util_test.go
```diff
@@ -1,11 +1,185 @@
...
+	"time"
 
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func TestToPtr(t *testing.T) {
+	intVar := 2
+	require.Equal(t, &intVar, ToPtr(intVar))
+
```
jaiveerk, line 15: can we add one for bool as well

aamgayle (approved):

jaiveerk (approved):
lgtm w one change/question

pkg/util/util_test.go
OliverMKing: no, bool is gone

PR #189, opened 2024-04-02
add default nic configuration as cli args

OliverMKing:
# Description

Exposes default nic configuration options through cli args.

## Type of change

# How Has This Been Tested?

Locally, unit, e2e

# Checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,6 +26,7 @@ var Flags = &Config{}
 var dnsZonesString string
 
 func init() {
+	flag.Var(&Flags.DefaultController, "default-controller", "kind of default controller to use. should be one of 'standard', 'public', 'private', or 'off'.")
 	flag.StringVar(&Flags.NS, "namespace", DefaultNs, "namespace for managed resources")
 	flag.StringVar(&Flags.Registry, "registry", "mcr.microsoft.com", "container image registry to use for managed components")
 	flag.StringVar(&Flags.MSIClientID, "msi", "", "client ID of the MSI to use when accessing Azure resources")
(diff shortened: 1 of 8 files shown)


pkg/config/config.go
```diff
@@ -26,6 +26,7 @@ var Flags = &Config{}
 var dnsZonesString string
 
 func init() {
+	flag.Var(&Flags.DefaultController, "default-controller", "kind of default controller to use. should be one of 'standard', 'public', 'private', or 'off'.")
```
bfoley13, line 29: PRD only has 3 options of  external/internal/none, what is the standard option?

OliverMKing: the current live experience of not reconciling annotations. It makes it so we can decide when we want to opt into this without having to cut a new operator release (and can roll back if needed).

OliverMKing: Without standard we wouldn't be able to get this in until we had the new api experience ready and we wouldn't be able to bring the new api experience to older versions of the operator

bfoley13: add the description of these options into the PR description with the implications of public/private overwriting changes. otherwise looks good!

bfoley13 (approved):

pkg/config/types.go
```diff
@@ -0,0 +1,79 @@
...
+func (c *ControllerConfig) String() string {
+	if c == nil {
+		return "nil"
+	}
+
+	if str, ok := controllerConfigMapping[*c]; ok {
+		return str
+	}
+
+	return "unknown"
```
jaiveerk, line 40: do we want to return standard here instead of a new "unknown" type if we can't recognize the enum? that seems like the safer path

OliverMKing: this is just the stringify method. returning an unknown string is appropriate. we properly return an error in the parsing logic

PR #194, opened 2024-04-15
add security context to externaldns

OliverMKing:
# Description

adds security context to external dns

### DIFF
--- a/pkg/manifests/external_dns.go
+++ b/pkg/manifests/external_dns.go
@@ -272,6 +272,17 @@ func newExternalDNSDeployment(conf *config.Config, externalDnsConfig *ExternalDn
 								corev1.ResourceMemory: resource.MustParse("250Mi"),
 							},
 						},
+						SecurityContext: &corev1.SecurityContext{
+							Privileged:               util.ToPtr(false),
+							AllowPrivilegeEscalation: util.ToPtr(false),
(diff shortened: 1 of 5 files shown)


Tatsinnit (approved):

pkg/manifests/external_dns.go
```diff
@@ -272,6 +272,17 @@ func newExternalDNSDeployment(conf *config.Config, externalDnsConfig *ExternalDn
 								corev1.ResourceMemory: resource.MustParse("250Mi"),
 							},
 						},
+						SecurityContext: &corev1.SecurityContext{
+							Privileged:               util.ToPtr(false),
+							AllowPrivilegeEscalation: util.ToPtr(false),
+							ReadOnlyRootFilesystem:   util.ToPtr(true),
+							RunAsNonRoot:             util.ToPtr(true),
+							RunAsUser:                util.Int64Ptr(65532),
```
jaiveerk, line 280: is this specific uid documented by externalDNS anywhere or is it arbitrary? wonder if there's a specific one we should use for the same reasons that we discussed during the operator securitycontext PR

jaiveerk (approved):

pkg/manifests/external_dns.go
OliverMKing: this is what's recommended by the external dns team. it's a generic non root

PR #196, opened 2024-04-16
bump to external dns 0.13.5

OliverMKing:
# Description

bumps us to external dns 0.13.5.

Need to land on this version before jumping to higher versions because of [this](https://github.com/kubernetes-sigs/external-dns/issues/3876). This is [documented by ExternalDNS](https://github.com/kubernetes-sigs/external-dns/releases/tag/v0.13.6).

We handle [this](https://github.com/kubernetes-sigs/external-dns/issues/2922) breaking change made by ExternalDNS that will come into affect as part of this bump. We mitigate the change by using the `--txt-wildcard-replacement=approutingwildcard` flag. This makes the chance of a collision all but impossible.

I also add e2e tests to cover the above wildcard case to prove it works.

### DIFF
--- a/testing/e2e/infra/provision.go
+++ b/testing/e2e/infra/provision.go
@@ -105,19 +105,19 @@ func (i *infra) Provision(ctx context.Context, tenantId, subscriptionId, applica
 		// that will change the loop variable capture to be the standard way loops work.
 		func(idx int) {
 			resEg.Go(func() error {
-				z, err := clients.NewZone(ctx, subscriptionId, i.ResourceGroup, fmt.Sprintf("zone-%d-%s", idx, i.Suffix))
+				z, err := clients.NewZone(ctx, subscriptionId, i.ResourceGroup, fmt.Sprintf("Zone-%d-%s", idx, i.Suffix))
 				if err != nil {
-					return logger.Error(lgr, fmt.Errorf("creating zone: %w", err))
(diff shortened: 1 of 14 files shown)


aamgayle (approved):
LGTM

testing/e2e/infra/provision.go
```diff
@@ -108,23 +106,45 @@ func (i *infra) Provision(ctx context.Context, tenantId, subscriptionId, applica
...
+
+				ret.Zones = append(ret.Zones, WithCert[Zone]{
+					Zone: z,
+					Cert: cert,
+				})
 				return nil
 			})
 		}(idx)
 	}
 	for idx := 0; idx < lenPrivateZones; idx++ {
```
jaiveerk, line 129: nit: why not have public and private zone creation happen, then block on kv creation, then block on kv creation for certs once? it requires another couple of for loops for private/public certs but allows more to happen while kv creation is going on in the background

jaiveerk (changes requested):
minor change requested but looks good otherwise

testing/e2e/infra/provision.go
OliverMKing: How would that allow more to happen while KV creation is blocking? The current pattern performs all possible work and only blocks once we require KV.

jaiveerk: you're so right -- missed the fact that the internal call is a goroutine

jaiveerk (approved):

PR #213, opened 2024-06-05
bump to go 1.22

OliverMKing:
# Description

bumps us to Go 1.22. Needed for a Gatekeeper integration for scanning our manifests.

Also update a few unit tests to reflect updated controller runtime packages. Fake client no longer fully infers types.


tbarnes94 (approved):

pkg/controller/service/ingress_reconciler_test.go
```diff
@@ -35,6 +35,8 @@ func TestIngressReconcilerIntegration(t *testing.T) {
 	svc := &corev1.Service{}
 	svc.UID = "test-svc-uid"
 	svc.Name = "test-service"
+	svc.APIVersion = "v1"
+	svc.Kind = "Service"
```
Vidya2606, line 39: for my understanding: what is the use of adding apiversion and kind here ?

OliverMKing: It's just a difference in how the fake client works due to the version bump.

bfoley13 (approved):

PR #259, opened 2024-07-22
Small fix for DnsZoneCountOne in basic e2e tests

aamgayle:
# Description
While testing other e2e tests I noticed that basic's tests did not create any zoners when using a single zone (DnsZoneCountOne) and thus skipped the test's logic and passed. Since it used DnsZoneCountOne instead of DnsZoneCountNone, the logic to catch a test that didn't have DNS zones also wasn't fired. Changing the variable name for the slice of zoners fixes the issues and the e2e tests for single zone tests still pass after testing. 

Fixes # (issue)
Feature # (details)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Unit Tests
- [X] e2e tests

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/testing/e2e/suites/basic.go
+++ b/testing/e2e/suites/basic.go
@@ -111,11 +111,11 @@ var clientServerTest = func(ctx context.Context, config *rest.Config, operator m
 	switch operator.Zones.Public {
 	case manifests.DnsZoneCountNone:
 	case manifests.DnsZoneCountOne:
-		zoners, err := toZoners(ctx, c, namespaces, infra.Zones[0])
+		zs, err := toZoners(ctx, c, namespaces, infra.Zones[0])
 		if err != nil {
 			return fmt.Errorf("converting to zoners: %w", err)
 		}
-		zoners = append(zoners, zoners...)
+		zoners = append(zoners, zs...)
 	case manifests.DnsZoneCountMultiple:
 		for _, z := range infra.Zones {
 			zs, err := toZoners(ctx, c, namespaces, z)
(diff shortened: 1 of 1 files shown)


OliverMKing (approved):

bosesuneha (approved):

jaiveerk (approved):

testing/e2e/suites/basic.go
```diff
@@ -111,11 +111,11 @@ var clientServerTest = func(ctx context.Context, config *rest.Config, operator m
 	switch operator.Zones.Public {
 	case manifests.DnsZoneCountNone:
 	case manifests.DnsZoneCountOne:
-		zoners, err := toZoners(ctx, c, namespaces, infra.Zones[0])
+		zs, err := toZoners(ctx, c, namespaces, infra.Zones[0])
```
Tatsinnit, line 114: ☕️💭 thanks for kind ping!! Please correct me but all changes are the name change of the variable? ❤️🙏

aamgayle: the issue was that zoners was being redeclared in and only in the DnsZoneCountOne's context so it was appending the values from infra.Zones[0] to the same slice that had those values and then did nothing with it afterward since those values disappeared outside of the context. 

Changing the variable name makes it so the only zoners slice that's there is the one that was originally declared above the switch and therefore has the values appended to it correctly. It's the reason the ones on DnsZoneCountMultiple worked this whole time. I attached an example of what happened here as well

image
image

PR #279, opened 2024-08-29
update trivy scanning to target active releases and nginx/externaldns dependencies

OliverMKing:
# Description

Updates our trivy scanning to run for all active releases. Also runs for nginx and externaldns versions tied to that release.

See an example run [here](https://github.com/OliverMKing/aks-app-routing/actions/runs/10622604514)


.github/workflows/scan.yml
```diff
@@ -0,0 +1,34 @@
...
+  gen-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+      - id: set-matrix
+        run: |
+          matrix=$(cat active_releases.json | tr '\n' ' ')
```
jaiveerk, line 20: Could you reduce this to a single workflow with an output for the operator, nginx, and externalDNS that you grab via jq in-line? Then pass those through as inputs directly to each trivvy step, where each step "needs" the checkout/parse step

jaiveerk (dismissed):

.github/workflows/scan.yml
OliverMKing: i could but then they wouldn't run in parallel and the fail strategy wouldn't make sense. what would be the advantage of what you're proposing?

jaiveerk: ah - didn't realize we can't run steps in parallel.

Seems like the community is pretty solidly in favor of this functionality too - https://github.com/orgs/community/discussions/26291

The proposed benefit would just be simplification of the code required to do this, but given the performance hit I'm not sure that the simplicity is worth it.

jaiveerk (approved):

PR #282, opened 2024-09-03
update e2e to use new operator versions

OliverMKing:
# Description

updates e2e to use new operator versions

### DIFF
--- a/testing/e2e/manifests/operator.go
+++ b/testing/e2e/manifests/operator.go
@@ -287,13 +287,6 @@ func Operator(latestImage string, publicZones, privateZones []string, cfg *Opera
 
 	// edit and select relevant manifest config by version
 	switch cfg.Version {
-	case OperatorVersion0_0_3:
-		baseDeployment.Spec.Template.Spec.Containers[0].ReadinessProbe = nil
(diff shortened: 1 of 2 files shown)


bfoley13 (approved):

testing/e2e/manifests/operator.go
```diff
@@ -39,19 +39,19 @@ var (
...
 	OperatorVersionLatest = math.MaxUint // this must always be the last/largest value in the enum because we order by value
 )
 
 func (o OperatorVersion) String() string {
 	switch o {
-	case OperatorVersion0_0_3:
-		return "0.0.3"
-	case OperatorVersion0_2_0:
-		return "0.2.0"
+	case OperatorVersion0_2_1_Patch_4:
```
jaiveerk, line 51: approving but might be worth looking into a way to automate which versions are used so that we don't have to update them every time we release a new patch/a new version we need to test.

jaiveerk (approved):

davidgamero (approved):
lgtm

testing/e2e/manifests/operator.go
OliverMKing: probably some way to utilize the `active_releases.json` file I added last week

PR #284, opened 2024-09-11
fix sec vulns and backport osm fixes

OliverMKing:
# Description

fixes sec vulns in v0.2.1 and backports osm fixes from https://github.com/Azure/aks-app-routing-operator/pull/170


Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM golang:1.20 as builder
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.22 as builder
 
 WORKDIR /go/src/github.com/Azure/aks-app-routing-operator
 ADD . .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -a -ldflags '-extldflags "-static"' -o aks-app-routing-operator
 
-FROM scratch
+FROM gcr.io/distroless/static:nonroot
```
jaiveerk, line 7: should we be using a microsoft distroless image?

jaiveerk (approved):

Dockerfile
OliverMKing: none of the other aks oss projects use anything other than distroless. I believe the microsoft distroless are internal only so they don't work with oss.

We can follow up on this but gcr is fine for now

bfoley13 (approved):

PR #306, opened 2024-10-28
Support for TLS Functionality with Gateway API via Workload Identity

jaiveerk:
# Description

This PR adds support for a managed TLS experience with the Gateway API via workload identity.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?
Unit tests. e2e tests will be added after DNS functionality has been merged.

# Checklist:


pkg/controller/controller.go
```diff
@@ -210,6 +212,17 @@ func setupControllers(mgr ctrl.Manager, conf *config.Config, lgr logr.Logger, cl
...
 	}
 
+	if conf.EnableGateway {
+		lgr.Info("setting up gateway reconcilers")
+		if err := keyvault.NewGatewaySecretClassProviderReconciler(mgr, conf); err != nil {
+			return fmt.Errorf("setting up Gateway SPC reconciler: %s", err)
+		}
+
+		if err := keyvault.NewKvServiceAccountReconciler(mgr); err != nil {
+			return fmt.Errorf("setting up Keyvault TLS default ServiceAccount reconciler: %s", err)
```
OliverMKing, line 222: %v not %s

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+var kvSaControllerName = controllername.New("gateway", "keyvault", "serviceaccount")
+
+// KvServiceAccountReconciler reconciles the default "azure-app-routing-kv" ServiceAccount in each namespace where users
```
OliverMKing, line 23: can users federate the credential with a subject of a non existing service account and have this work?

1. user federate non existing credential in namespace
2. user creates gateway object and this controller makes service account

does the federation properly apply to the newly created SA? Does it apply instantly or is there a wait time?

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+		For(&gatewayv1.Gateway{}), mgr.GetLogger()).
+		Complete(&KvServiceAccountReconciler{
+			client: mgr.GetClient(),
+			events: mgr.GetEventRecorderFor("app-routing-operator"),
+		})
+
+}
+
+func (k *KvServiceAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	res := ctrl.Result{}
```
OliverMKing, line 44: I know this is the pattern some other controllers use but I don't really like it. Named returns are better and accomplish the desired defer behavior better

OliverMKing: to clarify named return is better because you can follow go conventions instead of always having to remember to use the res and err vars. It's easy to mess up the current pattern because it relies on always using these exact vars

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+
+	logger, err := logr.FromContext(ctx)
+	if err != nil {
+		return res, err
+	}
+	logger = kvSaControllerName.AddToLogger(logger).WithValues("name", req.Name, "namespace", req.Namespace)
+
+	gwObj := &gatewayv1.Gateway{}
+	if err = k.client.Get(ctx, req.NamespacedName, gwObj); err != nil {
+		return res, client.IgnoreNotFound(err)
```
OliverMKing, line 59: this is the perfect example of why named return is better than the current metrics/err pattern. This will record as an error in metrics

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	clientId, err := extractClientIdForManagedSa(gwObj)
+	if clientId == "" {
+		return res, err
+	}
+
+	if err != nil {
```
OliverMKing, line 67: 63 should be triggered first so we never enter this block

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+	logger.Info("checking for existing ServiceAccount")
+
+	existing := &corev1.ServiceAccount{}
+	err = k.client.Get(ctx, types.NamespacedName{Name: toCreate.Name, Namespace: toCreate.Namespace}, existing)
+	if client.IgnoreNotFound(err) != nil {
+		logger.Error(err, "failed to fetch existing app routing serviceaccount")
+		return res, err
+	}
+
+	if existing != nil && existing.Annotations != nil && existing.Annotations[wiSaClientIdAnnotation] != "" {
```
OliverMKing, line 98: existing cannot be nil. you initialize it on 91

pkg/controller/keyvault/gateway_kv_serviceaccount.go
OliverMKing: I'm not a huge fan of the pattern of passing in client id through gateway and using that for the SA. It's a lot of indirection. Why doesn't user just create SA and pass that in instead? Drawing the line at us making SA but still taking in client id makes the entire process convoluted because gateway objects are many to one for namespaces and you can also have many tls configs in a single gateway object

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+		existingClientId := existing.Annotations[wiSaClientIdAnnotation]
+		if existingClientId != clientId {
+			errText := fmt.Sprintf("gateway specifies clientId %s but azure-app-routing-kv ServiceAccount already uses clientId %s", clientId, existingClientId)
```
OliverMKing, line 101: so it's impossible for them to change the client id for the service account? that doesn't seem correct. They technically could if they deleted the SA but that's a workaround

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,137 @@
...
+			logger.Info(errText)
+			k.events.Event(gwObj, Warning.String(), "InvalidInput", errText)
+			return res, err
+		}
+	}
+
+	if toCreate.Annotations == nil {
```
OliverMKing, line 108: when would this block not be nil? why not just set it to a map literal?

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,244 @@
...
+// and Keyvault URI in its TLS options field. The SPC is used to mirror the Keyvault values into
+// a k8s secret so that it can be used by the CRD controller.
+type GatewaySecretProviderClassReconciler struct {
+	client client.Client
+	events record.EventRecorder
+	config *config.Config
+}
+
+func NewGatewaySecretClassProviderReconciler(manager ctrl.Manager, conf *config.Config) error {
+	if conf.DisableKeyvault {
```
OliverMKing, line 43: inconsistent pattern to return early here for this but not for gateway disabled. we should stick to the existing pattern

pkg/controller/controller.go
jaiveerk: do you mean %w

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: They could modify the ServiceAccount first, right? We don't have any controllers watching the ServiceAccount itself

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: oh yeah good call - we shouldn't have this controller registered at all if Gateway isn't enabled.

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: yeah good call I'll move over to named returns

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: yeah good catch - realized my tests just check for the value of the error but miss capturing the event. I'll update accordingly

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: good catch - force of habit :D

pkg/controller/controller.go
OliverMKing: yes. missed the extra arch vv

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,244 @@
...
+			For(&gatewayv1.Gateway{}), manager.GetLogger(),
+	).Complete(&GatewaySecretProviderClassReconciler{
+		client: manager.GetClient(),
+		events: manager.GetEventRecorderFor("aks-app-routing-operator"),
+		config: conf,
+	})
+}
+
+func (g *GatewaySecretProviderClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	res := ctrl.Result{}
```
OliverMKing, line 60: change this pattern for all new controllers

jaiveerk: yeah will do! expanding pod placeholder tests then pushing a commit addressing all the above

pkg/controller/controller.go
jaiveerk: %vv == %w 👍 🚀

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,249 @@
...
+	// retrieve gateway resource from request + log the get attempt, but ignore not found
+	gwObj := &gatewayv1.Gateway{}
+	err = g.client.Get(ctx, req.NamespacedName, gwObj)
+
+	if err != nil {
+		err = client.IgnoreNotFound(err)
+		return
+	}
+
+	if gwObj.Spec.GatewayClassName != istioGatewayClassName {
```
OliverMKing, line 81: I don't know if we should have this. I'm cool with just working with all Gateway objects out of the box. No real reason to limit it other than we can't feasibly test everything but that's fine as long as our contracts are clear.

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,249 @@
...
+			For(&gatewayv1.Gateway{}), manager.GetLogger(),
+	).Complete(&GatewaySecretProviderClassReconciler{
+		client: manager.GetClient(),
+		events: manager.GetEventRecorderFor("aks-app-routing-operator"),
+		config: conf,
+	})
+}
+
+func (g *GatewaySecretProviderClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
+	res = ctrl.Result{}
```
OliverMKing, line 57: what's the point of these?

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// set up metrics given result/error
+	defer func() {
+		metrics.HandleControllerReconcileMetrics(gatewaySecretProviderControllerName, res, err)
+	}()
+
+	// set up logger
+	logger, err := logr.FromContext(ctx)
+	if err != nil {
+		return
```
OliverMKing, line 68: just return everything normally. This is no different from what you were doing early. Fully take advantage of the named return.

return ctrl.Result{}, fmt.Errorf("getting logr from context: "%w", err)

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,249 @@
...
+			NewControllerManagedBy(manager).
+			For(&gatewayv1.Gateway{}), manager.GetLogger(),
+	).Complete(&GatewaySecretProviderClassReconciler{
+		client: manager.GetClient(),
+		events: manager.GetEventRecorderFor("aks-app-routing-operator"),
+		config: conf,
+	})
+}
+
+func (g *GatewaySecretProviderClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
```
OliverMKing, line 56: i'd rename err to reterr or something so it's more obvious without overlap

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,249 @@
...
+					Controller: util.ToPtr(true),
+					Kind:       gwObj.Kind,
+					Name:       gwObj.Name,
+					UID:        gwObj.UID,
+				}},
+			},
+		}
+		logger = logger.WithValues("spc", spc.Name)
+
+		if listenerIsKvEnabled(listener) {
```
OliverMKing, line 107: use go conventions of breaking early if it's not enabled instead of continously nesting

pkg/controller/keyvault/kv_util.go
```diff
@@ -16,40 +15,60 @@ import (
 
 var nginxNamePrefix = "keyvault-nginx-"
 
-func buildSPC(obj client.Object, spc *secv1.SecretProviderClass, config *config.Config) (bool, error) {
-	var certURI, specSecretName string
+type EventType int
```
OliverMKing, line 18: where is this used?

pkg/controller/keyvault/kv_util.go
```diff
@@ -16,40 +15,60 @@ import (
...
+
+type SPCConfig struct {
+	ClientId        string
+	TenantId        string
+	KeyvaultCertUri string
+	Name            string
+	Cloud           string
+}
+
+func shouldDeploySpc(obj client.Object) bool {
```
OliverMKing, line 42: this abstraction doesn't work for gateway?

pkg/util/util.go
```diff
@@ -101,3 +101,28 @@ func FilterMap[K comparable, V comparable](m map[K]V, keep func(K, V) bool) map[
 
 	return new
 }
+
+// RequeueError is a specific kind of error that can be returned by operations that need to eventually be requeued following some duration
+type RequeueError struct {
```
OliverMKing, line 106: Not sure this abstraction makes sense. It's the callees responsibility to determine the retry time where you're using it here. I don't think this is needed

pkg/controller/keyvault/gateway_test_consts.go
```diff
@@ -0,0 +1,427 @@
+package keyvault
```
OliverMKing, line 1: not positive what the convention for naming a test helper should be but I suspect it should have the _test suffix

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: I just feel like it opens us up to mistakes/issues that might arise with users' misconfigurations of other ingress providers. Enforcing that we only work with Istio makes the managed experience we offer as narrow and clearly defined as possible which I think is going to be important for this.

@sabbour for extra thoughts on the UX side of things though - I could go either way

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: good catch, removing

pkg/controller/keyvault/kv_util.go
jaiveerk: I'd rather just have the constant than hardcode the string in multiple places to prevent typos/invalid event types. Here's a summary of the usage though:
image

pkg/controller/keyvault/kv_util.go
jaiveerk: No since SPCs in Gateways map to individual listeners, which don't satisfy the obj interface.

Rather than have one method that does two things, determine if an SPC should be deployed AND build the SPC itself, I figured it would make sense to break that up into two separate methods, which is what I do for a per-listener basis on each Gateway resource we reconcile. 

This way, we can keep the process of building SPCs the same, whether the SPC is based on an individual listener or on an ingress/nic - the SPCConfig object provides an abstraction for all three to work

pkg/util/util.go
jaiveerk: I thought about that, but I felt like it still makes more sense for something performing validation/verification to be the entity responsible for dictating whether or not a requeue is required. I could have such a method return a ctrl.Result{} with a requeue time in addition to the error but given that we already differentiate user errors from normal errors in validation, I figured this best matches that pattern.

pkg/controller/keyvault/gateway_test_consts.go
jaiveerk: updated!

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: updated

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: done

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: done

pkg/controller/keyvault/kv_util.go
OliverMKing: this isn't in the k8s libraries anywhere?

jaiveerk: It wasn't in the events library but I spent some time digging through the code and found it - replacing now, thanks for pointing this out

pkg/controller/controllername/controllername.go
```diff
@@ -28,12 +28,12 @@ type ControllerNamer interface {
 	AddToController(blder *builder.Builder, l logr.Logger) *builder.Builder
 }
 
-// controllerName ex. {"My","Controller", "Name"} ->  MyControllerName
-type controllerName []string
+// ControllerName ex. {"My","Controller", "Name"} ->  MyControllerName
```
OliverMKing, line 31: why are we exporting this?

OliverMKing: having it private was purposeful to force them to use New making it difficult to misues

jaiveerk: It helps keep our tests concise - check out the changes I made in the placeholder pod tests.

how would you feel about making it a separate struct containing a private string list? That would allow us to achieve the same

jaiveerk: actually wait I think I can work around this

jaiveerk: ok reverted to private with a bit of a workaround

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,134 @@
...
+		existingClientId := existing.Annotations[wiSaClientIdAnnotation]
+		if existingClientId != clientId {
+			errText := fmt.Sprintf("gateway specifies clientId %s but azure-app-routing-kv ServiceAccount already uses clientId %s", clientId, existingClientId)
+			logger.Info(errText)
+			k.events.Event(gwObj, corev1.EventTypeWarning, "InvalidInput", errText)
+			return ctrl.Result{}, nil
+		}
+	}
+
+	toCreate.Annotations = map[string]string{wiSaClientIdAnnotation: clientId}
```
OliverMKing, line 106: why not just put this where you define toCreate?

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	for _, listener := range gwObj.Spec.Listeners {
+		if listener.TLS == nil || listener.TLS.Options == nil {
+			continue
+		}
+		if listener.TLS.Options[clientIdTLSOption] != "" {
+			if ret != "" && string(listener.TLS.Options[clientIdTLSOption]) != ret {
+				return "", newUserError(fmt.Errorf("user specified multiple clientIds in one gateway resource"), "multiple unique clientIds specified. Please select one clientId to associate with the azure-app-routing-kv ServiceAccount")
+			}
+			ret = string(listener.TLS.Options[clientIdTLSOption])
```
OliverMKing, line 129: you don't have to index in twice. cleaner to just assign.

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+	err = g.client.Get(ctx, req.NamespacedName, gwObj)
+	if err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			logger.Error(err, "failed to fetch Gateway")
+			return ctrl.Result{}, fmt.Errorf("fetching gateway: %w", err)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if gwObj.Spec.GatewayClassName != istioGatewayClassName {
```
OliverMKing, line 77: We have this same logic a few other places. We should abstract to a single function so we can change our rules for when to allow for a gateway object to be reconciled easily

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+			var clientId string
+			clientId, err = retrieveClientIdFromListener(ctx, g.client, req.Namespace, listener.TLS.Options)
+			if err != nil {
+				var userErr userError
+				if errors.As(err, &userErr) {
+					logger.Info(fmt.Sprintf("failed to fetch clientId for SPC for listener %s due to user error: %s, sending warning event", listener.Name, userErr.userMessage))
+					g.events.Eventf(gwObj, corev1.EventTypeWarning, "InvalidInput", "invalid TLS configuration: %s", userErr.userMessage)
+					return ctrl.Result{}, nil
+				}
+				logger.Error(err, fmt.Sprintf("failed to fetch clientId for listener %s: %q", listener.Name, err.Error()))
```
OliverMKing, line 113: why %q?

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: yeah good call that's cleaner

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: left over from when I was quoting errs but forgot to switch this back - updating!

pkg/controller/keyvault/consts.go
```diff
@@ -3,4 +3,14 @@ package keyvault
 const (
 	tlsCertKvUriAnnotation   = "kubernetes.azure.com/tls-cert-keyvault-uri"
 	tlsCertManagedAnnotation = "kubernetes.azure.com/tls-cert-keyvault-managed"
+
+	istioGatewayClassName = "istio"
```
bfoley13, line 7: is this the gateway class name that istio always uses is this what the istio just happens to use?

pkg/controller/keyvault/gateway_kv_serviceaccount.go
bfoley13: I can vouch that you can create federated credentials without having a SA present in the cluster.

pkg/controller/keyvault/gateway_kv_serviceaccount.go
bfoley13: Not sure I like this either, can lead to confusion if I edit a SA but not the gateway. Any reason why you're forcing the SA edit over allowing it to control the SA from the gateway regardless?

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+	return listener.TLS != nil && listener.TLS.Options != nil && listener.TLS.Options[tlsCertKvUriAnnotation] != ""
+}
+
+func retrieveClientIdFromListener(ctx context.Context, k8sclient client.Client, namespace string, options map[gatewayv1.AnnotationKey]gatewayv1.AnnotationValue) (string, error) {
+	certUri := string(options[certUriTLSOption])
+	saName := string(options[serviceAccountTLSOption])
+	inputClientId := string(options[clientIdTLSOption])
+
+	// validate user input
+	if certUri != "" && (saName == "" && inputClientId == "") {
```
bfoley13, line 212: should this not be `||`?

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: I just think that since there's a many-to-one relatinonship between listeners and the app-routing service account that uses the clientId from that listener, it makes sense to have the source of truth be the "one" and not the "many"

pkg/controller/keyvault/gateway_secret_provider_class.go
bfoley13: ah just klidding its one or the other

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: Totally forgot to follow up on this but yes - confirmed via manual testing. It uses the oidc URL to specify the cluster, and then from what I can tell there's no validation against the ns or SA that's specified - it seems like portal just takes care of formatting the string for the oidc cred

pkg/controller/keyvault/gateway_kv_serviceaccount.go
OliverMKing: Or we could avoid this problem entirely by having user specify SA instead of specify client ID. I still think that's the better solution as it's more flexible and the behavior is straightforward

pkg/controller/keyvault/gateway_kv_serviceaccount.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	return ctrl.Result{}, util.Upsert(ctx, k.client, toCreate)
+}
+
+func extractClientIdForManagedSa(gwObj *gatewayv1.Gateway) (string, error) {
+	var ret string
+
+	if !shouldReconcileGateway(gwObj) {
```
OliverMKing, line 113: having this logic here doesn't make sense. functions should have one purpose. 

Purpose of this is extracting client id, not verifying if we should reconcile

pkg/controller/keyvault/consts.go
jaiveerk: It's the gateway class used by istio - see here: https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#configuring-a-gateway

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: how do we know which client ID customers want to use if they don't want to make their own serviceaccount? This was a specific ask from the PM side so I'm hesitant to just discard the managed SA option. I also just think that we should do whatever we can to make this experience as minimally intensive as possible given the number of steps required to set up WI

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+		fullErr := fmt.Errorf("failed to reconcile Gateway resource %s: %w", req.Name, err)
+		logger.Error(err, fullErr.Error())
+		g.events.Event(gwObj, corev1.EventTypeWarning, "FailedUpdateOrCreateGateway", fullErr.Error())
+		return ctrl.Result{}, fullErr
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func GenerateGwListenerCertName(gw string, listener gatewayv1.SectionName) string {
```
OliverMKing, line 193: make this private

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+	template := fmt.Sprintf("kv-gw-cert-%s-%s", gw, string(listener))
```
OliverMKing, line 194: template doesn't describe what this is

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: yeah that's a great point - that should be the first thing to happen in reconciliation

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+
+	var ret string
+	var err error
+	switch inputClientId {
+	case "":
+		// pull service account
+		wiSa := &corev1.ServiceAccount{}
+		err = k8sclient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: saName}, wiSa)
+		if err != nil {
+			if client.IgnoreNotFound(err) == nil {
```
OliverMKing, line 235: just check if it's the not found error. weird pattern to compare to nil instead of just making the actual comparison you want

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,252 @@
...
+				return "", newUserError(fmt.Errorf("user-specified serviceAccount %s does not exist", saName), fmt.Sprintf("serviceAccount %s does not exist", saName))
+			}
+			return "", fmt.Errorf("fetching serviceAccount %s: %s", saName, err)
+		}
+
+		if wiSa.Annotations == nil || wiSa.Annotations[wiSaClientIdAnnotation] == "" {
+			errString := fmt.Sprintf("workload identity MSI client ID must be specified for serviceAccount %s with annotation %s", saName, wiSaClientIdAnnotation)
+			return "", newUserError(errors.New("user-specified service account doesn't contain annotation with clientId"), errString)
+		}
+		ret = wiSa.Annotations[wiSaClientIdAnnotation]
```
OliverMKing, line 245: immediately return instead of assigning

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: renamed to certname

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: good call yeah that's cleaner

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -124,15 +129,28 @@ func (p *PlaceholderPodController) placeholderPodCleanCheck(obj client.Object) (
 		if t.Name == "" || t.Spec.IngressClassName == nil || !managed {
 			return true, nil
 		}
+	case *gatewayv1.Gateway:
+		if t.Spec.GatewayClassName != istioGatewayClassName {
```
OliverMKing, line 133: use the abstraction you made for this

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -180,17 +204,41 @@ func (p *PlaceholderPodController) reconcileObjectDeployment(dep *appsv1.Deploym
 		return result, nil
 	}
 
+	// Verify ServiceAccount exists (if Gateway)
```
OliverMKing, line 207: why are we verifying this? seems unnecessary for this logic to need to confirm that it exists. The Pod itself will check that and it's another reconciler that creates the SA

pkg/controller/keyvault/placeholder_pod.go
```diff
@@ -291,3 +341,66 @@ func (p *PlaceholderPodController) buildDeployment(ctx context.Context, dep *app
...
+// that the ServiceAccount reconciler completes before the SPC reconciler and the placeholder pod reconciler. As a result, if the ServiceAccount reconciler
+// has not finished creating the azure-app-routing-kv ServiceAccount, the operation should be re-queued.
+// This also ensures that the user has created the ServiceAccount referenced within their own annotations before we attempt (and potentially fail) to create
+// the placeholder pod.
+func (p *PlaceholderPodController) verifyServiceAccount(ctx context.Context, spc *secv1.SecretProviderClass, obj client.Object, logger logr.Logger) (string, error) {
+	var serviceAccount string
+
+	switch t := obj.(type) {
+	case *gatewayv1.Gateway:
+		logger.Info("verifying service account referenced by listener exists")
```
OliverMKing, line 356: this entire block is duplicated from spc code. We should use common functions

pkg/controller/keyvault/gateway_kv_serviceaccount.go
OliverMKing: We don't. The customer brings their own service account. It's not a worse experience, it's better because the behavior is well defined and it's more flexible.

Asking a customer to bring their own service account is not a large ask in any shape and makes the behavior more straightforward. Removing a single simple touch step isn't worth added complexity for a feature that should be straightforward

jaiveerk: I'll set up time with the PMs to talk this through -- this should've been brought up during the design and review process.

bfoley13: What did we resolve to here? Customers bring SA for all gateway related tasks off the bat?

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,240 @@
...
+		// we should delete the SPC if it exists
+		logger.Info(fmt.Sprintf("attempting to remove unused SPC %s", spc.Name))
+
+		deletionSpc := &secv1.SecretProviderClass{}
+		if err = g.client.Get(ctx, client.ObjectKeyFromObject(spc), deletionSpc); err != nil {
+			if client.IgnoreNotFound(err) != nil {
+				logger.Error(err, fmt.Sprintf("failed to fetch SPC for deletion %s", spc.Name))
+				return ctrl.Result{}, fmt.Errorf("fetching SPC for deletion: %w", err)
+			}
+			return ctrl.Result{}, nil
```
bfoley13, line 165: In the event there are listeners that are listenerIsKvEnabled and !listenerIsKvEnabled, you will return here without `Upserting` the gateway object with updated certificate refs. This might be a `continue` instead of a return.

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		if manifests.HasTopLevelLabels(deletionSpc.Labels) {
+			// return if we fail to delete, but otherwise, keep going
+			if err = g.client.Delete(ctx, deletionSpc); err != nil {
+				if client.IgnoreNotFound(err) != nil {
+					logger.Error(err, fmt.Sprintf("failed to delete SPC %s", spc.Name))
+					return ctrl.Result{}, fmt.Errorf("deleting SPC: %w", err)
+				}
+				return ctrl.Result{}, nil
```
bfoley13, line 175: same here as above

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: great catch! fixed + added a UT for this

pkg/controller/keyvault/placeholder_pod.go
jaiveerk: second point is no longer valid but would also add that I'd want to ensure we're checking that the SA is enabled with workload identity, which we do here

pkg/controller/keyvault/gateway_kv_serviceaccount.go
jaiveerk: yep exactly - resolving conversation accordingly

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,232 @@
...
+				ClientId:        clientId,
+				TenantId:        g.config.TenantID,
+				KeyvaultCertUri: certUri,
+				Name:            generateGwListenerCertName(gwObj.Name, listener.Name),
+			}
+			err = buildSPC(spc, spcConf)
+			if err != nil {
+				var userErr userError
+				if errors.As(err, &userErr) {
+					logger.Info("failed to build SecretProviderClass from user error: %q sending warning event", userErr.userMessage)
```
OliverMKing, line 130: %q typo

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,232 @@
...
+					Name:       gwObj.Name,
+					UID:        gwObj.UID,
+				}},
+			},
+		}
+		logger = logger.WithValues("spc", spc.Name)
+
+		if listenerIsKvEnabled(listener) {
+			var clientId string
+			clientId, err = retrieveClientIdFromListener(ctx, g.client, req.Namespace, listener.TLS.Options)
```
OliverMKing, line 104: the client id should come from the SA.

We need some way of triggering a reconcile if the SA changes their client id too

pkg/controller/keyvault/gateway_index.go
```diff
@@ -0,0 +1,70 @@
...
+
+func gatewayServiceAccountIndexFn(object client.Object) []string {
+	gateway, ok := object.(*gatewayv1.Gateway)
+	if !ok {
+		return nil
+	}
+
+	saSet := map[string]struct{}{}
+
+	for _, listener := range gateway.Spec.Listeners {
```
OliverMKing, line 30: need to consider namespaces

jaiveerk: we consider namespaces when requeuing requests for Gateways - see line 60.

All we can do within this function is pull names from the Gateway resources - verification via the client naturally happens when we're grabbing the actual gateways tied to each serviceaccount

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: done via sa indexer

config/gatewaycrd/v1.2.0.yaml
```diff
@@ -0,0 +1,10345 @@
+# Copyright 2024 The Kubernetes Authors.
```
OliverMKing, line 1: maybe this should be in the config/crd folder? "config/crd/gateway"

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+
+	}
+
+	logger.Info("reconciling Gateway resource with new secret refs for each TLS-enabled listener")
+	if err = util.Upsert(ctx, g.client, gwObj); err != nil {
+		fullErr := fmt.Errorf("failed to reconcile Gateway resource %s: %w", req.Name, err)
+		logger.Error(err, fullErr.Error())
+		g.events.Event(gwObj, corev1.EventTypeWarning, "FailedUpdateOrCreateGateway", fullErr.Error())
```
OliverMKing, line 187: This should maybe just be a pure update instead of an upsert.

If a user deletes a gateway we don't want to recreate it

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,231 @@
...
+	return listener.TLS != nil && listener.TLS.Options != nil && listener.TLS.Options[tlsCertKvUriAnnotation] != ""
+}
+
+func retrieveClientIdForListener(ctx context.Context, k8sclient client.Client, namespace string, options map[gatewayv1.AnnotationKey]gatewayv1.AnnotationValue) (string, error) {
+	certUri := string(options[certUriTLSOption])
+	saName := string(options[serviceAccountTLSOption])
+
+	// validate user input
+	if certUri != "" && saName == "" {
+		return "", newUserError(errors.New("user specified cert URI but no ServiceAccount in a listener"), "detected Keyvault Cert URI, but no ServiceAccount was provided")
```
OliverMKing, line 213: detected seems like the wrong word here. Tell the user exactly what they need to do to resolve the problem.

"KeyVault Cert URI provided but the required ServiceAccount option was not. Both are required for the KeyVault TLS experience Provide the ServiceAccount by doing x"

pkg/controller/keyvault/kv_util.go
```diff
@@ -1,55 +1,63 @@
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 	secv1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
 )
 
 var nginxNamePrefix = "keyvault-nginx-"
 
-func buildSPC(obj client.Object, spc *secv1.SecretProviderClass, config *config.Config) (bool, error) {
-	var certURI, specSecretName string
+type SPCConfig struct {
```
OliverMKing, line 23: this should be private

pkg/controller/keyvault/kv_util.go
```diff
@@ -137,3 +145,28 @@ func (b userError) UserError() string {
...
+	return gwObj.Spec.GatewayClassName == istioGatewayClassName
+}
+
+func GetServiceAccountAndVerifyWorkloadIdentity(ctx context.Context, k8sclient client.Client, saName, saNamespace string) (string, error) {
+	// ensure referenced serviceaccount exists
+	saObj := &corev1.ServiceAccount{}
+	err := k8sclient.Get(ctx, types.NamespacedName{Name: saName, Namespace: saNamespace}, saObj)
+
+	if client.IgnoreNotFound(err) != nil {
+		return "", err
```
OliverMKing, line 159: wrap error with information about what we were doing

config/gatewaycrd/v1.2.0.yaml
OliverMKing: where is this used? I don't see this used anywhere in this code

jaiveerk: it's required to create our mock client for tests - see https://github.com/Azure/aks-app-routing-operator/pull/306/files#diff-63440271bd8dc852878ea83b0c4d954a51f8e05348014b61c2f1bfbb16bc99a4R131

pkg/controller/keyvault/gateway_secret_provider_class.go
jaiveerk: yeah good point - didn't realize controller runtime watches for delete events

jaiveerk: although - if the gateway isn't found we exit early anyway - but I suppose it doesn't hurt in case the delete takes too long to reflect/we somehow end up finding it successfully at the beginning

OliverMKing: what I'm worried about is the race condition. The gateway is updated then they delete it in the middle but we recreate it here

jaiveerk: yeah valid - updated to use update

pkg/controller/keyvault/gateway_index.go
```diff
@@ -0,0 +1,70 @@
...
+
+func gatewayServiceAccountIndexFn(object client.Object) []string {
+	gateway, ok := object.(*gatewayv1.Gateway)
+	if !ok {
+		return nil
+	}
+
+	if gateway == nil {
+		return nil
+	}
```
Vidya2606, line 31: we can probably combine both 
if gateway == nil || !ok {
       return nil
}

jaiveerk: great catch! just added it in

pkg/controller/keyvault/gateway_secret_provider_class.go
```diff
@@ -0,0 +1,231 @@
...
+				return ctrl.Result{}, fmt.Errorf("building spc: %w", err)
+			}
+
+			logger.Info(fmt.Sprintf("reconciling SecretProviderClass %s for listener %s", spc.Name, listener.Name))
+			if err = util.Upsert(ctx, g.client, spc); err != nil {
+				fullErr := fmt.Errorf("failed to reconcile SecretProviderClass %s: %w", req.Name, err)
+				logger.Error(err, fullErr.Error())
+				g.events.Event(gwObj, corev1.EventTypeWarning, "FailedUpdateOrCreateSPC", fullErr.Error())
+				return ctrl.Result{}, fullErr
+			}
```
Vidya2606, line 146: consider creating a helper function for repeated util.Upsert calls to improve code reusability!

jaiveerk: I might be missing something but what're you referring to? We only call upsert once here for the secretproviderclass, and then a regular client update below (see the thread)

Vidya2606 (approved):
Great work! the changes look good :)

OliverMKing (approved):

PR #308, opened 2024-11-01
ExternalDNS Resource Creation Refactor

jaiveerk:
# Description

Refactoring how we create ExternalDNS resources to parameterize serviceaccount, namespace, and identity. It will be necessary to not couple these to the cluster conf fed in at application start time to allow for externaldns resource creation via crds/to allow for workload identity support.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested? 
Tests have been adjusted.

- [ ] Test A
- [ ] Test B

# Checklist:


pkg/manifests/common.go
```diff
@@ -52,14 +52,14 @@ func GetOwnerRefs(owner client.Object, controller bool) []metav1.OwnerReference
 	}}
 }
 
-func Namespace(conf *config.Config) *corev1.Namespace {
+func Namespace(conf *config.Config, nsName string) *corev1.Namespace {
```
OliverMKing, line 55: what is the point of this?

jaiveerk: decouples creation of a namespace object from the namespace specified in the conf. This'll be necessary when an overlay ns is specified in the externaldns crd

pkg/controller/dns/external_dns.go
```diff
@@ -147,14 +155,12 @@ func getLabels(instances ...instance) map[string]string {
...
 func cleanObjs(instances []instance) []cleanObj {
 	var cleanObjs []cleanObj
-	for _, instance := range instances {
-		if instance.action == clean {
-			obj := cleanObj{
-				resources: instance.resources,
-				labels:    getLabels(instance),
-			}
-			cleanObjs = append(cleanObjs, obj)
+	for _, instance := range filterAction(instances, clean) {
```
OliverMKing, line 158: why not just have this be 

cleanObjs = filterAction(instances, clean)?

pkg/manifests/external_dns.go
```diff
@@ -54,41 +87,50 @@ func (p Provider) String() string {
...
+type ExternalDnsConfig struct {
+	TenantId, Subscription, ResourceGroup,
+	ClientId, ServiceAccountName, Namespace string
+	IdentityType       IdentityType
+	ResourceType       ResourceType
+	Provider           Provider
+	DnsZoneResourceIDs []string
+}
+
+func externalDNSResourceName(e *ExternalDnsConfig) string {
```
OliverMKing, line 100: why not make this a method?

pkg/manifests/external_dns.go
```diff
@@ -143,53 +187,90 @@ func newExternalDNSClusterRole(conf *config.Config, externalDnsConfig *ExternalD
...
-			},
 			{
 				APIGroups: []string{""},
 				Resources: []string{"nodes"},
 				Verbs:     []string{"get", "watch", "list"},
 			},
 		},
 	}
+
+	switch externalDnsConfig.ResourceType {
```
OliverMKing, line 198: why not make this logic a method on your new config type?

pkg/manifests/external_dns.go
```diff
@@ -143,53 +187,90 @@ func newExternalDNSClusterRole(conf *config.Config, externalDnsConfig *ExternalD
...
+
+	return role
 }
 
 func newExternalDNSClusterRoleBinding(conf *config.Config, externalDnsConfig *ExternalDnsConfig) *rbacv1.ClusterRoleBinding {
-	return &rbacv1.ClusterRoleBinding{
+	var serviceAccount string
+	switch externalDnsConfig.IdentityType {
+	case IdentityTypeWorkloadIdentity:
+		serviceAccount = externalDnsConfig.ServiceAccountName
```
OliverMKing, line 230: this should probably be logic on your new config type. We are mixing responsibilities here and splitting them between this new config type and these functions

pkg/manifests/common.go
jaiveerk: We shouldn't infer namespace name from the config anyway - there's nothing inherently about the need to create a namespace object that implies that it must be the one specified in the conf.

If you'd like I can make a second "namespace" method for only externalDNS resources to use (and not nginx ingress controller pods) but that seems like overkill.

pkg/controller/dns/external_dns.go
jaiveerk: we need to extract the cleanobjs type from each instance, so we first need to filter to look for the clean instances, then create cleanobjs from each one.

cleanObjs = filterAction(instances, clean) would yield a list of instances, not cleanObjs

pkg/manifests/external_dns.go
jaiveerk: good call - this can be made into a method I'll adjust accordingly

pkg/manifests/external_dns.go
jaiveerk: you mean make it a method on the resource type?

I think it makes more sense to be left as a separate function - the rbac isn't as much a property or action of the externaldns config as it is a set of resources that need to be generated according to state in it (hence function vs method). If you feel strongly I can refactor, but doesn't seem necessary or flexible, especially if roles need to change based on other objects/parameters in the future

pkg/manifests/external_dns_test.go
```diff
@@ -26,6 +26,9 @@ var (
 		TenantId:           "test-tenant-id",
 		Subscription:       "test-subscription-id",
 		ResourceGroup:      "test-resource-group-public",
+		Namespace:          "test-namespace",
+		IdentityType:       IdentityTypeMSI,
+		ResourceType:       ResourceTypeIngress,
```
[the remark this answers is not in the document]
jaiveerk: done

pkg/manifests/external_dns.go
```diff
@@ -54,41 +87,50 @@ func (p Provider) String() string {
...
 }
 
-func (p Provider) ResourceName() string {
-	switch p {
+// ExternalDnsConfig defines configuration options for required resources for external dns
+type ExternalDnsConfig struct {
+	TenantId, Subscription, ResourceGroup,
+	ClientId, ServiceAccountName, Namespace string
+	IdentityType       IdentityType
+	ResourceType       ResourceType
```
OliverMKing, line 95: we should support multiple resource types

pkg/manifests/external_dns.go
jaiveerk: done, made it an external function.

pkg/manifests/external_dns.go
jaiveerk: I see now what you mean, took this out to a separate method (addResourceSpecificRules)

pkg/manifests/external_dns.go
jaiveerk: done

pkg/manifests/external_dns.go
```diff
@@ -54,41 +86,45 @@ func (p Provider) String() string {
...
 	var objs []client.Object
-
-	// Can safely assume the namespace exists if using kube-system
-	if conf.NS != "kube-system" {
-		objs = append(objs, Namespace(conf))
-	}
-
 	for _, dnsConfig := range externalDnsConfigs {
+		// Can safely assume the namespace exists if using kube-system
+		if dnsConfig.Namespace != "" && dnsConfig.Namespace != "kube-system" {
```
OliverMKing, line 125: where do we default to a namespace in dnsConfig?

we the default should be conf.NS (that's the old behavior)

pkg/manifests/external_dns.go
```diff
@@ -54,41 +86,45 @@ func (p Provider) String() string {
...
-		return externalDnsResourceName + "-private"
-	default:
-		return ""
+// ExternalDnsConfig defines configuration options for required resources for external dns
+type ExternalDnsConfig struct {
+	TenantId, Subscription, ResourceGroup,
+	ClientId, ServiceAccountName, Namespace,
+	CRDName string
+	IdentityType       IdentityType
+	ResourceTypes      []ResourceType
```
OliverMKing, line 95: seems fine but would be better to make this a set to avoid any potential for misuse

OliverMKing (dismissed):

pkg/manifests/external_dns.go
OliverMKing: we should really control the constructor for the externalDns configs within this package to prevent misuse

jaiveerk: I agree with the use of the constructor - i'll make the type private and add a public constructor, so that there's no need to default on a namespace

pkg/controller/dns/external_dns.go
```diff
@@ -91,24 +91,37 @@ func instances(conf *config.Config) []instance {
...
-	return &manifests.ExternalDnsConfig{
-		TenantId:           conf.TenantID,
-		Subscription:       conf.PublicZoneConfig.Subscription,
-		ResourceGroup:      conf.PublicZoneConfig.ResourceGroup,
-		Provider:           manifests.PublicProvider,
-		DnsZoneResourceIDs: util.Keys(conf.PublicZoneConfig.ZoneIds),
-	}
+func publicConfigForIngress(conf *config.Config) *manifests.ExternalDnsConfig {
+	return manifests.NewExternalDNSConfig(
+		conf,
```
OliverMKing, line 96: we are passing in so many arguments here. should use a struct as input instead, it's too long.

pkg/manifests/external_dns.go
```diff
@@ -54,41 +86,86 @@ func (p Provider) String() string {
...
 }
 
-// ExternalDnsResources returns Kubernetes objects required for external dns
-func ExternalDnsResources(conf *config.Config, externalDnsConfigs []*ExternalDnsConfig) []client.Object {
-	var objs []client.Object
+func (e *ExternalDnsConfig) DnsZoneResourceIds() []string {
+	return e.dnsZoneResourceIDs
+}
+
+func NewExternalDNSConfig(conf *config.Config, tenantId, subscription, resourceGroup, clientId, serviceAccountName, namespace, crdName string, identityType IdentityType, resourceTypes []ResourceType, provider provider, dnszoneresourceids []string) *ExternalDnsConfig {
```
OliverMKing, line 117: the point of having a constructor is validating some input. we should validate input and return an error here so we can make assumptions about the ExternalDnsConfig

pkg/manifests/external_dns.go
```diff
@@ -297,3 +383,58 @@ func newExternalDNSDeployment(conf *config.Config, externalDnsConfig *ExternalDn
 		},
 	}
 }
+
+func getServiceAccount(externalDnsConfig *ExternalDnsConfig) string {
```
OliverMKing, line 387: method on externalDnsConfig instead?

jaiveerk: resolving since I moved this to constructor

pkg/manifests/external_dns.go
```diff
@@ -54,41 +86,86 @@ func (p Provider) String() string {
...
+		serviceAccountName: serviceAccountName,
+		namespace:          namespace,
+		crdName:            crdName,
+		identityType:       identityType,
+		resourceTypes:      resourceTypes,
+		provider:           provider,
+		dnsZoneResourceIDs: dnszoneresourceids,
+	}
+
+	ret.resources = externalDnsResources(conf, []*ExternalDnsConfig{ret})
```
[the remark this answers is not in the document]
jaiveerk: resolving per offline discussion

pkg/manifests/external_dns.go
jaiveerk: added validation

pkg/controller/dns/external_dns.go
jaiveerk: resolving per offline discussion re: correct usage of params

pkg/controller/dns/external_dns.go
```diff
@@ -88,27 +92,50 @@ func instances(conf *config.Config) []instance {
...
+		"",
+		conf.NS,
+		"",
+		manifests.IdentityTypeMSI,
+		[]manifests.ResourceType{manifests.ResourceTypeIngress},
+		manifests.PublicProvider,
+		util.Keys(conf.PublicZoneConfig.ZoneIds))
+
+	if err != nil {
+		// do something
```
OliverMKing, line 114: do something

pkg/manifests/external_dns.go
```diff
@@ -54,41 +87,121 @@ func (p Provider) String() string {
...
+	return e.dnsZoneResourceIDs
+}
+
+func NewExternalDNSConfig(conf *config.Config, tenantId, subscription, resourceGroup, clientId, inputServiceAccount, namespace, crdName string, identityType IdentityType, resourceTypes []ResourceType, provider provider, dnszoneresourceids []string) (*ExternalDnsConfig, error) {
+	// valid values for enums
+	if identityType != IdentityTypeMSI && identityType != IdentityTypeWorkloadIdentity {
+		return nil, fmt.Errorf("invalid identity type: %v", identityType)
+	}
+
+	containsGateway := false
```
OliverMKing, line 124: remove any crd logic. add that when we add crds

OliverMKing: doesn't make sense to make preemtive changes. crd changes might look different that what we want. not worth early optimizing

jaiveerk: done

pkg/controller/dns/external_dns.go
jaiveerk: did something :D sorry about this not sure how it got rolled back lol

pkg/controller/dns/external_dns.go
jaiveerk: so we'd have one externaldnsconfig that's visible to the controllers and one internally? that just seems like a confusing pattern.

Either we declare the type in the controller, which would create a cyclical dependency, or we expose the fields in the public type - but that violates the initial premise of forming the constructor (ie to prevent incorrect usage/assignment of the fields).

We could also create a third type within manifests but that seems excessive

OliverMKing (approved):

OliverMKing (dismissed):

pkg/manifests/external_dns_test.go
```diff
@@ -82,16 +138,208 @@ var (
...
+			inputResourceName:  "test-resource",
+			identityType:       IdentityTypeWorkloadIdentity,
+			resourceTypes:      []ResourceType{ResourceTypeIngress},
+			provider:           PrivateProvider,
+			dnsZoneResourceIDs: []string{privateZoneOne, privateZoneTwo},
+			expectedError:      errors.New("workload identity requires a service account name"),
+		},
+	}
+	for _, tc := range testCases {
+		ret, err := NewExternalDNSConfig(tc.conf, InputExternalDNSConfig{TenantId: tc.tenantId, Subscription: tc.subscription, ResourceGroup: tc.resourceGroup, ClientId: tc.msiclientID, InputServiceAccount: tc.serviceAccountName, Namespace: tc.namespace, InputResourceName: tc.inputResourceName, IdentityType: tc.identityType, ResourceTypes: tc.resourceTypes, Provider: tc.provider, DnsZoneresourceIDs: tc.dnsZoneResourceIDs})
```
davidgamero, line 334: if this test struct mostly contains fields that are populating this InputExternalDNSConfig couldn't we just make a struct of this type on the test case to avoid making another layer on top of it?

jaiveerk: done

davidgamero (approved):

OliverMKing (approved):

PR #320, opened 2024-12-11
Adding CRD Definition for ExternalDNS Configuration

jaiveerk:
# Description

Adding CRDs for users to configure ExternalDNS deployments

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


sabbour:
Why does Cluster External DNS Configuration use namespace for filters while External DNS Configuration uses labels?

jaiveerk:
@sabbour 
> Why does Cluster External DNS Configuration use namespace for filters while External DNS Configuration uses labels?

The namespace filter in ClusterExternalDNSConfiguration is for users to specify which namespaces the cluster-scoped resource should look for resources in. By default, it will check all.

ExternalDNS configuration just checks for HTTPRoute/Gateway resources within the namespace it's deployed in, so it's more narrowly scoped/doesn't need to have any namespace filters.

Both are able to filter via labels.

PROJECT
```diff
@@ -10,11 +10,26 @@ repo: github.com/Azure/aks-app-routing-operator
 resources:
 - api:
     crdVersion: v1
-    namespaced: false
```
OliverMKing, line 13: why was this removed?

PROJECT
```diff
@@ -10,11 +10,26 @@ repo: github.com/Azure/aks-app-routing-operator
...
   group: approuting
   kind: NginxIngressController
   path: github.com/Azure/aks-app-routing-operator/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: kubernetes.azure.com
+  group: approuting
+  kind: ExternalDNSConfiguration
```
OliverMKing, line 24: This name seems off. It's not an ExternalDNS configuration. It's literally ExternalDNS.

A Deployment isn't called DeploymentConfiguration, it's 1:1 with a Deployment

PROJECT
```diff
@@ -10,11 +10,26 @@ repo: github.com/Azure/aks-app-routing-operator
...
+  path: github.com/Azure/aks-app-routing-operator/api/v1alpha1
+  version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
```
OliverMKing, line 29: the global one wouldn't be namespaced

PROJECT
```diff
@@ -10,11 +10,26 @@ repo: github.com/Azure/aks-app-routing-operator
...
+  domain: kubernetes.azure.com
+  group: approuting
+  kind: ClusterExternalDNSConfiguration
```
OliverMKing, line 32: ClusterExternalDNS

api/v1alpha1/clusterexternaldnsconfiguration_types.go
```diff
@@ -0,0 +1,94 @@
...
+	// TenantID is the ID of the Azure tenant where the DNS zones are located.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Format:=uuid
+	// +kubebuilder:validation:Pattern=`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`
+	TenantID string `json:"tenantID"`
+
+	// DNSZoneResourceIDs is a list of Azure Resource IDs of the DNS zones that the ExternalDNS controller should manage. These should be in the same resource group and be of the same type (public or private).
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:validation:MaxItems:=20
```
OliverMKing, line 49: where does this come from?

OliverMKing: why 20?

api/v1alpha1/clusterexternaldnsconfiguration_types.go
```diff
@@ -0,0 +1,94 @@
...
+	// +kubebuilder:validation:items:Enum:=ingress;gateway
+	// +listType:=set
+	ResourceTypes []string `json:"resourceTypes"`
+
+	// Identity contains information about the identity that ExternalDNS will use to interface with Azure resources.
+	// +kubebuilder:validation:Required
+	Identity ExternalDNSConfigurationIdentity `json:"identity"`
+
+	// ResourceNamespace is the namespace where the ExternalDNS resources will be deployed by app routing.
+	// +kubebuilder:validation:Required
```
OliverMKing, line 66: this shouldn't be required and should just default to app-routing-system

OliverMKing: or maybe it should be since they need to bring the SA. Hmmm

api/v1alpha1/externaldnsconfiguration_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions"`
+
+	ExternalDNSReadyReplicas       int32 `json:"externalDNSReadyReplicas"`
+	ExternalDNSUnavailableReplicas int32 `json:"externalDNSUnavailableReplicas"`
+
+	// Count of hash collisions for the managed resources. The App Routing Operator uses this field
+	// as a collision avoidance mechanism when it needs to create the name for the managed resources.
+	// +optional
+	// +kubebuilder:validation:Maximum=5
```
OliverMKing, line 116: why a max of 5 here? We aren't providing a configurable name for ExternalDNS so I don't think it makes sense to cap at 5

PROJECT
jaiveerk: not sure this is weird - noticed it's also namespaceed=true for the cluster config despite setting it explicitly to be cluster scoped so I'll look at this

PROJECT
jaiveerk: this isn't 1:1 with a deployment though - it's a set of other resources as well that all configure the externaldns deployment (hence the "configuration")

api/v1alpha1/clusterexternaldnsconfiguration_types.go
jaiveerk: it's arbitrary at the moment. @sabbour do we have a better figure for what our users want?

Post unblocking of large testing framework I can also look to see how large we can make one deployment - externaldns doesn't document any limitations

api/v1alpha1/clusterexternaldnsconfiguration_types.go
jaiveerk: yeah the whole point of this was that we shouldn't deploy to app-routing-system because that would require users to edit resources there (and it'll be locked down going forward in AKS automatic clusters). hence they need to pick the namespace where their resources are deployed + where their SA will be

OliverMKing: seems fine to me then

PROJECT
OliverMKing: what other resources is it? It's effectively just ExternalDNS.

A Deployment isn't just a ReplicaSet. It's also a Pod. My point is that Kubernetes resources are named after what abstractions they are. This isn't a Configuration, it's literally ExternalDNS

api/v1alpha1/clusterexternaldnsconfiguration_types.go
OliverMKing: externaldns scales on records not domains. 20 seems fine for now but would probably be good to document this in comments for now

api/v1alpha1/externaldnsconfiguration_types.go
jaiveerk: resolving per offline discussion

api/v1alpha1/clusterexternaldns_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Format:=uuid
+	// +kubebuilder:validation:Pattern=`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`
+	TenantID string `json:"tenantID"`
+
+	// DNSZoneResourceIDs is a list of Azure Resource IDs of the DNS zones that the ExternalDNS controller should manage. These must be in the same resource group and be of the same type (public or private).
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:validation:MaxItems:=20
+	// +kubebuilder:validation:items:Pattern:=`(?i)\/subscriptions\/(.{36})\/resourcegroups\/(.+?)\/providers\/Microsoft.network\/(dnszones|privatednszones)\/(.+)`
```
OliverMKing, line 49: can we validate that all the resource groups / subs are the same with cel. might be possible

api/v1alpha1/clusterexternaldns_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// +kubebuilder:validation:Required
+	Identity ExternalDNSIdentity `json:"identity"`
+
+	// ResourceNamespace is the namespace where the ExternalDNS resources will be deployed by app routing.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9][-a-z0-9\.]*[a-z0-9]$`
+	ResourceNamespace string `json:"resourceNamespace"`
+
```
OliverMKing, line 70: weren't we going to add a name field here? Would be like https://github.com/Azure/aks-app-routing-operator/blob/212cdfb4a39d34fcaada109738411bef94438cc9/api/v1alpha1/nginxingresscontroller_types.go#L47C2-L47C22

api/v1alpha1/externaldns_types.go
```diff
@@ -0,0 +1,131 @@
...
+	Filters ExternalDNSFilters `json:"filters,omitempty"`
+}
+
+// ExternalDNSIdentity contains information about the identity that ExternalDNS will use to interface with Azure resources.
+type ExternalDNSIdentity struct {
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9][-a-z0-9\.]*[a-z0-9]$`
+
+	// ServiceAccount is the name of the Kubernetes ServiceAccount that ExternalDNS will use to interface with Azure resources. It must be in the same namespace as the ExternalDNS.
```
OliverMKing, line 86: required

OliverMKing (dismissed):

api/v1alpha1/clusterexternaldns_types.go
jaiveerk: I was thinking it'd be easier/more intuitive to have that in the controller + emit events if not

api/v1alpha1/clusterexternaldns_types.go
jaiveerk: was planning on using the resourcename itself vs having multiple places to specify it

api/v1alpha1/externaldns_types.go
jaiveerk: good catch on that!

api/v1alpha1/clusterexternaldns_types.go
```diff
@@ -0,0 +1,92 @@
...
+// ClusterExternalDNS allows users to specify desired the state of a cluster-scoped ExternalDNS deployment and includes information about the state of their resources in the form of Kubernetes events.
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,shortName=cedns
+type ClusterExternalDNS struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ClusterExternalDNSSpec   `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
jaiveerk: done!

api/v1alpha1/clusterexternaldns_types.go
OliverMKing: we should have another field. That's more flexible. They might have a naming convention in their cluster to stick to

api/v1alpha1/clusterexternaldns_types.go
OliverMKing: it's better at this level is possible because it blocks the resource from being even created and why. Eventing is an okay mechanism for reporting information but for invalid specs it's much much better to block at the create level

api/v1alpha1/clusterexternaldns_types.go
```diff
@@ -0,0 +1,107 @@
...
+	// +kubebuilder:validation:MaxItems:=7
+	// +kubebuilder:validation:items:UniqueItems:=true
+	// +kubebuilder:validation:items:MaxLength:=172
+	// +kubebuilder:validation:items:MinLength:=103
+	// +kubebuilder:validation:items:MaxProperties:=1
+	// +kubebuilder:validation:XValidation:rule="self.all(item, item.split('/')[2] == self[0].split('/')[2])",message="all items must have the same subscription ID"
+	// +kubebuilder:validation:XValidation:rule="self.all(item, item.split('/')[4] == self[0].split('/')[4])",message="all items must have the same resource group"
+	// +kubebuilder:validation:XValidation:rule="self.all(item, item.split('/')[7] == self[0].split('/')[7])",message="all items must be of the same resource type"
+	// +listType:=set
+	DNSZoneResourceIDs []string `json:"dnsZoneResourceIDs"`
```
OliverMKing, line 65: need regex validation that it's a resource Id form?

jaiveerk: validating via kubebuilder:validation:items:pattern doesn't work - it doesn't actually end up appearing on the final resource definition that gets generated. Tried to use XValidation instead but the rule is too complex even with the lowered number of maximum permitted items.

api/v1alpha1/clusterexternaldns_types.go
```diff
@@ -0,0 +1,107 @@
...
+	// +kubebuilder:validation:Format:=uuid
+	// +kubebuilder:validation:Pattern=`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`
+	TenantID string `json:"tenantID"`
+
+	// DNSZoneResourceIDs is a list of Azure Resource IDs of the DNS zones that the ExternalDNS controller should manage. These must be in the same resource group and be of the same type (public or private). The number of zones is currently capped at 20 but may be expanded in the future.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:validation:MaxItems:=7
+	// +kubebuilder:validation:items:UniqueItems:=true
+	// +kubebuilder:validation:items:MaxLength:=172
```
bfoley13, line 58: for my sanity whats the formula for these

api/v1alpha1/clusterexternaldns_types.go
bfoley13: Name or name prefix?

testing/e2e/suites/externaldns_crd.go
```diff
@@ -0,0 +1,330 @@
...
+				lgr := logger.FromContext(ctx)
+				lgr.Info("starting test")
+
+				tcs := []struct {
+					name          string
+					ed            *v1alpha1.ExternalDNS
+					expectedError error
+				}{
+					{
+						name:          "valid",
```
bfoley13, line 57: filters in some tests here as well

api/v1alpha1/externaldns_types.go
```diff
@@ -0,0 +1,146 @@
...
+type ExternalDNSIdentity struct {
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9][-a-z0-9\.]*[a-z0-9]$`
+	// +kubebuilder:validation:Required
+	// ServiceAccount is the name of the Kubernetes ServiceAccount that ExternalDNS will use to interface with Azure resources. It must be in the same namespace as the ExternalDNS.
+	ServiceAccount string `json:"serviceAccount"`
+}
+
+type ExternalDNSFilters struct {
```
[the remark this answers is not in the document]
jaiveerk: Yeah, if a user made an ingress/gateway corresponding to one of the domains/names of the dns zones provided to externaldns, it would be synced to the zone - so ingress creation it would need to be RBAC-controlled.

Users with concerns over unwanted ingresses/routing though would need to use namespaces as a security boundary and use the namespaced CRD. But, you're correct that if a user has the ability to create an ingress, then there's nothing that's stopping from externalDNS publishing a record for that ingress to the zone. It would then be up to the cluster admin to ensure that only the platform/infra team has the ability to create ingresses, but if they believe that that's application-team scoped, then yeah anyone on the application team would be able to create an ingress.

testing/e2e/suites/clusterexternaldns_crd.go
```diff
@@ -0,0 +1,353 @@
...
+				lgr := logger.FromContext(ctx)
+				lgr.Info("starting test")
+
+				tcs := []struct {
+					name          string
+					ced           *v1alpha1.ClusterExternalDNS
+					expectedError error
+				}{
+					{
+						name:          "valid",
```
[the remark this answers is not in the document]
jaiveerk: filter doesn't have any restrictions on the value since it's optional, so there's nothing really to test from a CEL perspective. Filter functionality will be covered via integration tests and/or e2e tests that ensure that externaldns only reads gateway resources with the correct filters, and not all that exist in the cluster - but that can really only be tested once the controller for the CRD gets built out

testing/e2e/suites/externaldns_crd.go
jaiveerk: see above

jaiveerk:
ok-to-test sha=a5cf5b6

testing/e2e/suites/externaldns_crd.go
bfoley13: What am I looking for here? I dont see you added Filters to the `ExternalDNSSpec` for any test?

jaiveerk: Sorry, meant look above on the other comment you left about filters.

Filter doesn't have any restrictions on the value since it's optional, so there's nothing really to test from a CEL perspective. Filter functionality will be covered via integration tests and/or e2e tests that ensure that externaldns only reads gateway resources with the correct filters, and not all that exist in the cluster - but that can really only be tested once the controller for the CRD gets built out

api/v1alpha1/clusterexternaldns_types.go
```diff
@@ -0,0 +1,107 @@
...
+	// +kubebuilder:validation:Required
+	ResourceName string `json:"resourceName,omitempty"`
+
+	// TenantID is the ID of the Azure tenant where the DNS zones are located.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Format:=uuid
+	// +kubebuilder:validation:Pattern=`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`
+	TenantID string `json:"tenantID"`
+
+	// DNSZoneResourceIDs is a list of Azure Resource IDs of the DNS zones that the ExternalDNS controller should manage. These must be in the same resource group and be of the same type (public or private). The number of zones is currently capped at 20 but may be expanded in the future.
```
OliverMKing, line 53: comment is incorrect since we cap at 7

api/v1alpha1/externaldns_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// GatewayLabels contains key-value pairs that the ExternalDNS controller will use to filter the Gateways that it manages.
+	// +optional
+	GatewayLabels map[string]string `json:"gatewayLabels,omitempty"`
+
+	// RouteAndIngressLabels contains key-value pairs that the ExternalDNS controller will use to filter the HTTPRoutes and Ingresses that it manages.
+	// +optional
+	RouteAndIngressLabels map[string]string `json:"routeAndIngressLabels,omitempty"`
```
OliverMKing, line 112: why are route and ingress ones combined but gateway ones aren't?

jaiveerk: externaldns has a separate argument for gateway but uses the same argument for both ingresses and routes - 

gateway specific filter is documented [here](https://kubernetes-sigs.github.io/external-dns/v0.14.1/sources/gateway/#matching-gateways)

meanwhile, there's no mention of an ingress-label-filter on the [ingress page](https://kubernetes-sigs.github.io/external-dns/v0.14.1/sources/ingress/), and there's no page for routes on the sources page, but the [gateway tutorial](https://kubernetes-sigs.github.io/external-dns/v0.14.1/tutorials/gateway-api/#manifest-with-rbac) just comments that the "--label-filter" argument would apply to routes

OliverMKing (approved):

testing/e2e/suites/externaldns_crd.go
jaiveerk: When you say externaldns spec - are you referring to the deployment that gets created, or the actual CR?

If the former - yes, we do expect that config to eventually propagate to the deployment args. However, that would happen via the controller that we make for the ExternalDNS CR, but that will be created (and tested) in a separate PR once the CRD definitions are actually in.

If the latter - are you suggesting a UT that sets the filters in an instance of the CR in golang/code and then validates that those filters actually made it to the object that gets saved by the APIserver?

OliverMKing (approved):

testing/e2e/suites/externaldns_crd.go
bfoley13: More about the later but I see you're not doing a validation of the ExternalDNS resource after upsert

bfoley13 (approved):

testing/e2e/suites/externaldns_crd.go
jaiveerk: I feel like that's more of a test of the Upsert method itself and/or the apiserver's ability to correctly store the object but that will be validated anyway via tests over the controller's functionality, but appreciate the callout + will make sure that there are adequate tests once we get the controller up

PR #345, opened 2025-01-27
restructure repo into single go project with multiple cmd dir entrypoints

OliverMKing:
# Description

Restructures the repo into having a cmd directory with multiple entrypoints (just Operator and e2e for now). Also shifts everything to under the same go.mod. Was always annoying having multiple (need to update dependencies in multiple places).

This better matches how other mature Go projects are structured and is something we'll need for planned future additions (ManagedTLS will use new entrypoints).


.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-	"go.buildTags": "e2e"
+	"files.associations": {
+		// test files embedded in e2e with a main method, can't be suffixed with .go because Go will complain about main
```
jaiveerk, line 3: why not leave e2e as its own proj/main cmd in its directory?

jaiveerk, line 3: I think we should move away from this embedding approach entirely, and move towards what we have in RP.

If we're breaking down the project into a series of cmds anyway, why not create a directory for each kind of deployment we create and containerize them? That way, like we do in rp, we can leverage things like syntax highlighting etc, and potentially move towards more complex clients/servers in the future without keeping them contained in single files.

Feel like that's broadly a more flexible approach anyway

jaiveerk (changes requested):

OliverMKing:
Ci/cd will fail on this PR until it's merged in. Just how GitHub actions work since it's running the ones in main

.vscode/settings.json
jaiveerk: why not just do it here? this seems like the optimal time given that we're already rearranging their file structure, which is going to be the extent of the changes regardless

In other words - I'm not sure that this PR gets significantly more complex if we just make that change now vs leaving it as tech debt going forward

bfoley13 (approved):

jaiveerk (dismissed):

PR #365, opened 2025-02-13
add wait shutdown and delay to ingress-nginx controller

OliverMKing:
# Description

add wait shutdown preStop lifecycle hook to nginx and a grace period. Without this if a ingress-nginx pod is deleting the LoadBalancer service continues routing to the deleting ingress-nginx pod even when it's unable to serve traffic because of how health checks with the Azure Load Balancer work.


pkg/manifests/nginx.go
```diff
@@ -444,6 +444,13 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 						Name:  "controller",
 						Image: path.Join(conf.Registry, "/oss/kubernetes/ingress/nginx-ingress-controller:"+ingressConfig.Version.tag),
 						Args:  deploymentArgs,
+						Lifecycle: &corev1.Lifecycle{
+							PreStop: &corev1.LifecycleHandler{
+								Exec: &corev1.ExecAction{
+									Command: []string{"/wait-shutdown"},
```
jaiveerk, line 450: can we add a comment here explaining why this is relevant? feel like it's not super obvious, maybe even just a link to nginx docs

jaiveerk (dismissed):

jim-minter (changes requested):
Note: this was opened in response to a query from SIG-Workload, and we don't believe that this change will help with our issue.  Given that, I wouldn't merge this at this time.

cc @ThatJuanGuy

pkg/manifests/nginx.go
OliverMKing: documented

pkg/manifests/nginx.go
```diff
@@ -444,6 +448,14 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 						Name:  "controller",
 						Image: path.Join(conf.Registry, "/oss/kubernetes/ingress/nginx-ingress-controller:"+ingressConfig.Version.tag),
 						Args:  deploymentArgs,
+						Lifecycle: &corev1.Lifecycle{
```
jim-minter, line 451: I'm not sure about what the effect is of adding the preStop is and if/why it's necessary, but the other part LGTM.

OliverMKing: I read through the upstream code, issues, and PRs and it seems to do absolutely nothing. Removing

jaiveerk: waiting on removal + will approve after!

jim-minter (approved):
LGTM

jaiveerk (approved):

PR #373, opened 2025-02-27
bump pause image used in placeholder pod

OliverMKing:
# Description

bumps pause image used in placeholder pod


meecethereese (approved):

jaiveerk (approved):
pause, but newer!

PR #374, opened 2025-02-28
respect rollouts in ingress controller topology spread

OliverMKing:
# Description

respects rollouts in ingress controller topology spread. Without this, during new rollouts pods aren't necessarily spread out across all nodes because older versions of the deployment pods exist and interfere with calculations then are removed.

### DIFF
--- a/pkg/manifests/nginx.go
+++ b/pkg/manifests/nginx.go
@@ -437,6 +437,11 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 							TopologyKey:       "kubernetes.io/hostname", // spread across nodes
 							WhenUnsatisfiable: corev1.ScheduleAnyway,
 							LabelSelector:     selector,
+							MatchLabelKeys: []string{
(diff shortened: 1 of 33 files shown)


pkg/manifests/nginx.go
```diff
@@ -437,6 +437,11 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 							TopologyKey:       "kubernetes.io/hostname", // spread across nodes
 							WhenUnsatisfiable: corev1.ScheduleAnyway,
 							LabelSelector:     selector,
+							MatchLabelKeys: []string{
+								// https://kubernetes.io/blog/2024/08/16/matchlabelkeys-podaffinity/
```
OliverMKing, line 441: should only work for aks 1.31+. Should be fine just important to note

OliverMKing: it won't break on previous versions, just won't do anything

jim-minter (approved):
LGTM

jaiveerk (approved):

PR #376, opened 2025-03-06
Adding Controller for ExternalDNS CRD (ClusterExternalDNS coming soon)

jaiveerk:
# Description

Adding controller for ExternalDNS CRD to generate appropriate ExternalDNS resources.

Given the length of this PR already, I'm planning on including changes for the ClusterExternalDNS CRD resource in a separate PR that should be smaller, as it will leverage many of the changes defined in this one.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

- [x] fixture tests to ensure manifests generated from CRD 
- [x] unit tests to ensure controller performs proper validation for user input and generates correct resources

# Checklist:


pkg/controller/dns/external_dns_crd.go
```diff
@@ -0,0 +1,118 @@
...
+	// set up logger
+	logger, err := logr.FromContext(ctx)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("creating logger: %w", err)
+	}
+	logger = ExternalDNSCRDControllerName.AddToLogger(logger).WithValues("namespace", req.Namespace, "name", req.Name)
+
+	obj := &v1alpha1.ExternalDNS{}
+	if err = e.client.Get(ctx, req.NamespacedName, obj); err != nil {
+		logger.Error(client.IgnoreNotFound(err), "failed to get externaldns object, will ignore not found error")
```
OliverMKing, line 60: i like not logging at all if it's not found

pkg/controller/dns/external_dns_crd.go
```diff
@@ -0,0 +1,118 @@
...
+	manifestsConf, err := manifests.NewExternalDNSConfig(e.config, inputDNSConf)
+	if err != nil {
+		logger.Error(err, "failed to generate ExternalDNS resources from ExternalDNS CR")
+		e.events.Eventf(obj, corev1.EventTypeWarning, "FailedUpdateOrCreateExternalDNSResources", "failed to generate external DNS resources: %s", err)
+		return ctrl.Result{}, err
+	}
+
+	resources := manifestsConf.Resources()
+
+	multiError := &multierror.Error{ErrorFormat: func(errs []error) string {
```
OliverMKing, line 87: what's the reason for customizing the errorformat here? looks like the default one to me

pkg/controller/dns/external_dns_crd.go
```diff
@@ -0,0 +1,118 @@
...
+			Kind:       obj.Kind,
+			Name:       obj.Name,
+			UID:        obj.UID,
+		}})
+
+		currentResourceErr := util.Upsert(ctx, e.client, resource)
+		if currentResourceErr != nil {
+			logger.Error(currentResourceErr, "failed to upsert externaldns resources")
+		}
+		multiError = multierror.Append(multiError, currentResourceErr)
```
OliverMKing, line 108: i think this is an interesting pattern where we don't exit early and keep upserting resources if there's a failure. I think this is good but this is different from our other controllers

OliverMKing: My kneejerk opinion is that just exiting early is better because it's simpler but trying to think of the pros and cons. It's hard to think of a scenario where upserting all even though there's an error is helpful

pkg/controller/dns/external_dns_crd.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	if multiError.ErrorOrNil() != nil {
+		logger.Error(err, "failed to upsert externaldns resources")
+		e.events.Eventf(obj, corev1.EventTypeWarning, "FailedUpdateOrCreateExternalDNSResources", "failed to deploy external DNS resources: %s", multiError.Error())
+		return ctrl.Result{}, errors.New(multiError.Error())
+	}
+
```
OliverMKing, line 116: there's no logic for updating the Status fields of externaldns. I'm look with adding that in future revisions but just want to call that out so we don't forget about them

pkg/controller/dns/external_dns_crd_consts_for_test.go
```diff
@@ -0,0 +1,330 @@
+package dns
```
OliverMKing, line 1: nit: pretty verbose file names. think you can just simplify all of them

pkg/controller/dns/util_test.go
```diff
@@ -0,0 +1,13 @@
+package dns
+
+import (
+	"testing"
+)
+
```
OliverMKing, line 10: missing some tests haha

pkg/controller/keyvault/kv_util.go
```diff
@@ -127,23 +127,23 @@ func certSecretName(ingressName string) string {
 	return fmt.Sprintf("keyvault-%s", ingressName)
 }
 
-type userError struct {
+type UserError struct {
```
OliverMKing, line 130: don't really like exporting this from kv package. weird for this to be under kv package really. 

Move to utils where it makes sense? Or maybe a new errors pkg?

OliverMKing: if we export from kv it's just kinda weird and the more we do that the more chance of circular dependency

pkg/manifests/external_dns.go
```diff
@@ -126,12 +127,21 @@ func (p Provider) string() string {
...
+	// Provider is specified when an InputConfig is coming from the MC External DNS Reconciler, since no zones may be provided for the clean case
+	Provider *Provider
+	// IdentityType can either be MSI or WorkloadIdentity
+	IdentityType IdentityType
+	// ResourceTypes refer to the resource types that ExternalDNS should look for to configure DNS. These can include Gateway and/or Ingress
+	ResourceTypes map[ResourceType]struct{}
+	// DnsZoneresourceIDs contains the DNS zones that ExternalDNS will use to configure DNS
+	DnsZoneresourceIDs []string
+	// Filters contains various filters that ExternalDNS will use to filter resources it scans for DNS configuration
+	Filters *v1alpha1.ExternalDNSFilters
```
OliverMKing, line 142: I think this should be a new type instead of mixing crd and manifest domains.

pkg/manifests/external_dns.go
```diff
@@ -143,6 +153,12 @@ type ExternalDnsConfig struct {
 	identityType  IdentityType
 	resourceTypes map[ResourceType]struct{}
 	provider      Provider
+	isNamespaced  bool
+
+	// crd-specific specific fields
+	routeAndIngressLabelSelector string
+	gatewayLabelSelector         string
+	// TODO: USE THIS
```
OliverMKing, line 161: what's this mean?

pkg/controller/dns/external_dns_crd.go
jaiveerk: so I was thinking about this - should I just add extra lines to log if it's not a not-found error? I wouldn't want to end up in a situation where we can't pull our CRDs (for some weird reason) and we don't have immediate logging telling us why

pkg/controller/dns/external_dns_crd.go
jaiveerk: I think it would help diagnose the issue more quickly. If there's a partial failure, it's easier to pinpoint which resource failed to deploy, and the error then contains all failed resources instead of just the first one that hit failure

pkg/controller/dns/external_dns_crd.go
jaiveerk: you mean separate from the setcondition method on the crd? https://github.com/jaiveerk/aks-app-routing-operator/blob/3fbeb7179817c1fccdd04ee2c5853a65886ec28b/api/v1alpha1/externaldns_types.go#L45-L45

pkg/controller/dns/util_test.go
jaiveerk: added 😄 gotta love pushing my notes preemptively lol

pkg/manifests/external_dns.go
jaiveerk: I was thinking about that -- 

The biggest reason why I chose to put it here is because all other forms of validation for input take place here, whether it's for input from the legacy scenario or from CRDs. That being said, we could just have that check earlier on and then marshall it into a separate intermediate type, but I don't think that we gain anything by doing that. If anything, since the filters are unique to the CRD case anyway, I figured we may as well keep that type propagated through as far as we could and then have a fixed domain that we're operating in to add these labels.

What do you think?

pkg/manifests/external_dns.go
jaiveerk: left behind while coding lol, removed

pkg/controller/keyvault/kv_util.go
jaiveerk: resolving after refactor PR was merged in

pkg/controller/dns/external_dns_crd.go
```diff
@@ -0,0 +1,109 @@
...
+	config *config.Config
+	client client.Client
+	events record.EventRecorder
+}
+
+func NewExternalDNSCRDController(manager ctrl.Manager, config config.Config) error {
+	return ExternalDNSCRDControllerName.AddToController(ctrl.NewControllerManagedBy(manager).
+		For(&v1alpha1.ExternalDNS{}).
+		Owns(&appsv1.Deployment{}).
+		Owns(&corev1.ConfigMap{}), manager.GetLogger()).
```
OliverMKing, line 36: what's the point of owning the configmap? Just want to prevent user from changing it and bricking themselves?

jaiveerk: yep

pkg/controller/dns/external_dns_crd.go
OliverMKing: yes. What's the problem with just logging if it's not not found?

OliverMKing: It's extremely common for not found to be the case because controller runtime requeues on delete

jaiveerk: ok yeah sounds good I can have this operate in the same way our other ones do, I'll have it log if not found otherwise log/return error

pkg/controller/dns/external_dns_crd.go
OliverMKing: no. There's no logic to set those fields. This is the right controller to do that in

pkg/manifests/external_dns.go
```diff
@@ -77,21 +78,25 @@ func (rt ResourceType) generateResourceDeploymentArgs() []string {
...
-				Verbs:     []string{"get", "watch", "list"},
-			},
+		ret := []rbacv1.PolicyRule{
 			{
 				APIGroups: []string{"gateway.networking.k8s.io"},
 				Resources: []string{"gateways", "httproutes", "grpcroutes"},
 				Verbs:     []string{"get", "watch", "list"},
 			},
 		}
+		if !dnsconfig.isNamespaced {
```
OliverMKing, line 91: surprising that these are the only permissions that change

OliverMKing (approved):

pkg/manifests/external_dns.go
jaiveerk: yeah, the broader difference is role vs cluster role

OliverMKing (approved):

meecethereese (approved):

PR #377, opened 2025-03-07
Refactor User Error

jaiveerk:
Refactor user error to make accessible outside of kv package

### DIFF
--- a/pkg/controller/controllerutils/controller_errors.go
+++ b/pkg/controller/controllerutils/controller_errors.go
@@ -1,4 +1,4 @@
-package controllererrors
+package controllerutils
 
 type UserError struct {
 	Err         error
--- a/pkg/controller/controllerutils/serviceaccount.go
+++ b/pkg/controller/controllerutils/serviceaccount.go
@@ -1,0 +1,34 @@
+package controllerutils
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const wiSaClientIdAnnotation = "azure.workload.identity/client-id"
+
+func GetServiceAccountAndVerifyWorkloadIdentity(ctx context.Context, k8sclient client.Client, saName, saNamespace string) (string, error) {
+	// ensure referenced serviceaccount exists
+	saObj := &corev1.ServiceAccount{}
+	err := k8sclient.Get(ctx, types.NamespacedName{Name: saName, Namespace: saNamespace}, saObj)
+
+	if client.IgnoreNotFound(err) != nil {
+		return "", fmt.Errorf("failed to fetch serviceaccount to verify workload identity configuration: %s", err)
+	}
+
+	// SA wasn't found, return appropriate error
+	if err != nil {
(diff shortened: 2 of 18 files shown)


meecethereese (approved):

OliverMKing (dismissed):

pkg/controller/controllerutils/controller_errors.go
```diff
@@ -0,0 +1,20 @@
+package controllerutils
```
OliverMKing, line 1: honestly it probably just makes sense to move these to the utils package

OliverMKing: doesn't make sense for there to be a difference

jaiveerk: these are pretty scoped to controller logic though, right? we're never going to have a "user error" that gets created from other packages.

I like having this be within controller because going forward we'll likely want more controller-specific utils anyway

OliverMKing: same for upsert and some of the utils in the utils package

pkg/controller/controllerutils/serviceaccount.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const wiSaClientIdAnnotation = "azure.workload.identity/client-id"
+
+func GetServiceAccountAndVerifyWorkloadIdentity(ctx context.Context, k8sclient client.Client, saName, saNamespace string) (string, error) {
```
OliverMKing, line 15: need unit test

pkg/controller/controllerutils/controller_errors.go
jaiveerk: touche

jaiveerk: moved

pkg/controller/controllerutils/serviceaccount.go
jaiveerk: added

pkg/util/serviceaccount.go
```diff
@@ -0,0 +1,34 @@
...
+
+const wiSaClientIdAnnotation = "azure.workload.identity/client-id"
+
+func GetServiceAccountAndVerifyWorkloadIdentity(ctx context.Context, k8sclient client.Client, saName, saNamespace string) (string, error) {
+	// ensure referenced serviceaccount exists
+	saObj := &corev1.ServiceAccount{}
+	err := k8sclient.Get(ctx, types.NamespacedName{Name: saName, Namespace: saNamespace}, saObj)
+
+	if client.IgnoreNotFound(err) != nil {
+		return "", fmt.Errorf("failed to fetch serviceaccount to verify workload identity configuration: %s", err)
```
OliverMKing, line 21: nit: %w is ideal for wrapping errors

OliverMKing (dismissed):
LGTM

pkg/util/serviceaccount.go
jaiveerk: added

meecethereese (approved):

OliverMKing (approved):

PR #387, opened 2025-03-17
add ManagedCertificate crd

OliverMKing:
# Description

adds the ManagedCertificate crd


PROJECT
```diff
@@ -19,7 +18,6 @@ resources:
   version: v1alpha1
 - api:
     crdVersion: v1
-    namespaced: false
```
jaiveerk, line 22: is the removal of these intentional?

api/v1alpha1/managedcertificate_types.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Important: Run "make crd" to regenerate code after modifying this file
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ManagedCertificateSpec defines the desired state of ManagedCertificate.
+type ManagedCertificateSpec struct {
+	// Target defines the targets that the Certificate will be bound to.
```
jaiveerk, line 12: nit: target here, singular?

api/v1alpha1/managedcertificate_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// +kubebuilder:validation:Required
+	ServiceAccount string `json:"serviceAccount,omitempty"`
+}
+
+// ManagedCertificateTarget defines the targets that a Certificate will be bound to.
+// +kubebuilder:validation:MinProperties=1
+// +kubebuilder:validation:MaxProperties=1
+type ManagedCertificateTarget struct {
+	// Secret is the name of the Secret that will contain the Certificate.
+	Secret string `json:"secret,omitempty"`
```
jaiveerk, line 36: Is secret required?

api/v1alpha1/managedcertificate_types.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// ManagedCertificateDnsZone defines the DNS Zone that a ManagedCertificate will be applied to.
+type ManagedCertificateDnsZone struct {
+	// ResourceId is the Azure Resource ID of the DNS Zone. Can be retrieved with `az network dns zone show -g <resource-group> -n <zone-name> --query id -o tsv`.
+	ResourceId string `json:"resourceId,omitempty"`
```
jaiveerk, line 42: Can we add CEL validation here? See external dns crd

api/v1alpha1/managedcertificate_types.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// below fields are needed for cross-tenant scenarios
+
+	// TenantId is the Azure Tenant ID of the DNS Zone.
+	// +kubebuilder:validation:Optional
+	TenantId string `json:"tenantId,omitempty"`
```
jaiveerk, line 48: would be helpful to have cel validation here too

PROJECT
OliverMKing: just how kubebuilder did it. Doesn't have an affect on the code or crds so it's just a default thing for them

api/v1alpha1/managedcertificate_types.go
OliverMKing: Plural on purpose. There will be other targets in the future (mounting directly to Istio Ingress)

api/v1alpha1/managedcertificate_types.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// DnsZone defines the DNS Zone that the ManagedCertificate will be applied to.
+	// +kubebuilder:validation:Required
+	DnsZone ManagedCertificateDnsZone `json:"dnsZone,omitempty"`
+
+	// DomainNames is a list of domain names that the Certificate will be issued for.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +listType=set
+	DomainNames []string `json:"domainNames,omitempty"`
```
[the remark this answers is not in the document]
OliverMKing: Will add in a future PR. I like to add the CEL ones when I add e2e tests confirming the CEL. Need to add some logic to operator too before that
 
Don't want to make this pr too big

api/v1alpha1/managedcertificate_types.go
OliverMKing: done. good catch

api/v1alpha1/managedcertificate_types.go
```diff
@@ -0,0 +1,109 @@
...
+// ManagedCertificateDnsZone defines the DNS Zone that a ManagedCertificate will be applied to.
+type ManagedCertificateDnsZone struct {
+	// ResourceId is the Azure Resource ID of the DNS Zone. Can be retrieved with `az network dns zone show -g <resource-group> -n <zone-name> --query id -o tsv`.
+	ResourceId string `json:"resourceId,omitempty"`
+
+	// below fields are needed for cross-tenant scenarios
+
+	// TenantId is the Azure Tenant ID of the DNS Zone.
+	// +kubebuilder:validation:Optional
+	TenantId string `json:"tenantId,omitempty"`
```
jaiveerk, line 49: behavior on DNS CRDs is to have tenant ID as a top level, required field. Not saying one is correct vs the other, but we should likely have them match. Do you think we should have this be required, or have it be optional in both and use config's tenantID?

OliverMKing: it should default to the cluster's tenant in both

OliverMKing: most users shouldn't ever need this field

jaiveerk: alright, I'll put together a PR changing that for existing DNS resources once clusterDNS controller PR is merged

jaiveerk (dismissed):

api/v1alpha1/managedcertificate_types.go
OliverMKing: actually wait no. min and max properties enforce this. I tricked myself

OliverMKing: I did that on purpose for future extensibility when we support just mounting to istio ingress pod directly

PR #391, opened 2025-03-19
add disable expensive cache

OliverMKing:
# Description

adds disable expensive cache. Cache uses lots of memory on super large clusters for resources that app routing doesn't need to cache today.


pkg/config/config.go
```diff
@@ -46,6 +48,7 @@ func init() {
 	flag.DurationVar(&Flags.DnsSyncInterval, "dns-sync-interval", defaultDnsSyncInterval, "interval at which to sync DNS records")
 	flag.StringVar(&Flags.CrdPath, "crd", "/crd", "location of the CRD manifests. manifests should be directly in this directory, not in a subdirectory")
 	flag.BoolVar(&Flags.EnableGateway, "enable-gateway", false, "whether or not to support and create controllers for Gateway API resources")
+	flag.BoolVar(&Flags.DisableExpensiveCache, "disable-expensive-cache", false, "disable the cache for expensive resources that aren't core to App Routing like Pods and Events")
```
jaiveerk, line 51: don't we want this to be true by default?

OliverMKing: discussed in teams

pkg/controller/controller.go
```diff
@@ -84,6 +86,25 @@ func NewRestConfig(conf *config.Config) *rest.Config {
...
 
 func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager, error) {
+	// client creates ListWatch cache for all resources that are
+	// GET or LISTed. this uses too much memory on large clusters.
+	var clientOpts *client.CacheOptions
+	cacheOpts := cache.Options{}
+	if conf.DisableExpensiveCache {
+		clientOpts = &client.CacheOptions{
+			// we only interact with a small subset of pods
+			// but there's no good filter to only cache the pods we care about
```
jaiveerk, line 96: why not top-level filters for pods? don't we apply them to everything we deploy?

pkg/controller/keyvault/event_mirror.go
```diff
@@ -24,8 +25,19 @@ import (
...
+
+const (
+	involvedObjectKindField = "involvedObject.kind"
+	eventReasonFailedMount  = "FailedMount"
+	reasonField             = "reason"
+	eventKindPod            = "Pod"
+)
+
+// EventMirrorSelector is a selector for Events that are relevant to the EventMirror controller
+var EventMirrorSelector = fields.AndSelectors(
```
jaiveerk, line 38: more of a high level question that may not be relevant to this PR but why don't we also mirror events from SPC? isn't that what contains the aggregation for events from CSI?

meecethereese (approved):

jaiveerk (approved):

pkg/controller/keyvault/event_mirror.go
OliverMKing: I don't know. Would maybe make sense to mirror from SPC as well if there's useful info reflected there. Original owners of this probably determined that the Pod FailedMount error is most relevant though

We should probably mirror events to Gateway and NginxIngressController (in case of default certs) as well.

pkg/controller/controller.go
OliverMKing: the placeholder pod doesn't actually have top level labels (it's the deployment that does)

https://github.com/Azure/aks-app-routing-operator/blob/175e5e1fe8b429f9c49254a37be4ec53f0c60785/pkg/controller/keyvault/placeholder_pod.go#L284

Good idea for an improvement here though. I'll add that in a future iteration

PR #392, opened 2025-03-20
add initial proto tooling

OliverMKing:
# Description

Adds the initial proto tooling to this repo. Will be used to communicate with ManagedCertificate CX.

The example file is just to prove this works. Will be removed shortly in a subsequent PR but just proves it works for now.

We'll need some custom proto tooling like redact but those will be added in future PRs


proto/example/v1/example.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
```
OliverMKing, line 1: this will be removed shortly. Just proves it works right now.

meecethereese (approved):

jaiveerk (approved):

PR #421, opened 2025-04-25
shift from update to ssa patch in controllers

OliverMKing:
# Description

shift from update to ssa patch in controllers. This is more resilient and lets us fully take control over resources.


jaiveerk (approved):
lgtm

PR #427, opened 2025-04-29
add unit visualizations

OliverMKing:
# Description

adds unit test visualizations so we can see why our unit tests take time to run. This is helpful so I can parallelize some tests that are bottlenecks


jaiveerk (approved):
big fan

PR #434, opened 2025-05-12
add kubernetes.azure.com/set-kube-service-host-fqdn to pods

OliverMKing:
# Description

adds kubernetes.azure.com/set-kube-service-host-fqdn annotation to pods. https://learn.microsoft.com/en-us/azure/aks/outbound-rules-control-egress#required-outbound-network-rules-and-fqdns-for-aks-clusters

This helps prevent firewalls from interfering with our pods communicating with the api server.

### DIFF
--- a/pkg/controller/dns/consts_for_test.go
+++ b/pkg/controller/dns/consts_for_test.go
@@ -71,30 +71,32 @@ var clusterHappyPathPublic = &v1alpha1.ClusterExternalDNS{
 	},
 }
 
-var happyPathPublicJSON = `{"cloud":"","location":"","resourceGroup":"test-rg","subscriptionId":"12345678-1234-1234-1234-123456789012","tenantId":"12345678-1234-1234-1234-123456789012","useWorkloadIdentityExtension":true}`
-var happyPathPublicJSONHash = sha256.Sum256([]byte(happyPathPublicJSON))
-var happyPathPublicNoTenantIDJSON = `{"cloud":"","location":"","resourceGroup":"test-rg","subscriptionId":"12345678-1234-1234-1234-123456789012","tenantId":"12345678-1234-1234-1234-012987654321","useWorkloadIdentityExtension":true}`
-var happyPathPublicNoTenantIDJSONHash = sha256.Sum256([]byte(happyPathPublicNoTenantIDJSON))
-var happyPathPublicConfigmap = &corev1.ConfigMap{
-	TypeMeta: metav1.TypeMeta{
-		Kind:       "ConfigMap",
-		APIVersion: "v1",
-	},
-	ObjectMeta: metav1.ObjectMeta{
-		ResourceVersion: "1",
-		Name:            "happy-path-public-external-dns",
-		Namespace:       "test-ns",
-		Labels: map[string]string{
-			"app.kubernetes.io/managed-by":   "aks-app-routing-operator",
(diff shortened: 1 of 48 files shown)


pkg/controller/dns/consts_for_test.go
```diff
@@ -112,7 +114,8 @@ var happyPathPublicDeployment = &appsv1.Deployment{
 		Selector: &metav1.LabelSelector{MatchLabels: map[string]string{"app": "happy-path-public-external-dns"}},
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
-				Labels: map[string]string{"app.kubernetes.io/managed-by": "aks-app-routing-operator", "kubernetes.azure.com/managedby": "aks", "app": "happy-path-public-external-dns", "checksum/configmap": hex.EncodeToString(happyPathPublicJSONHash[:])[:16]},
+				Labels:      map[string]string{"app.kubernetes.io/managed-by": "aks-app-routing-operator", "kubernetes.azure.com/managedby": "aks", "app": "happy-path-public-external-dns", "checksum/configmap": hex.EncodeToString(happyPathPublicJSONHash[:])[:16]},
+				Annotations: map[string]string{"kubernetes.azure.com/set-kube-service-host-fqdn": "true"},
```
OliverMKing, line 118: only additions here is adding the annotations to the testing constants.

These tests should be improved. Manifests package is what's responsible for determining exact resources provisioned by app routing. When we make changes to manifests package it should be tested in the manifests package and we shouldn't need to change things here.

I'll refactor this in a followup PR. Nginx manifests export a type that allows you to get definition for specific resources (like Deployments, configmap, etc). We'll add this improvement to external dns manifests export too and consume them here rather than redefine what should be completely handled by manifest package.

pkg/manifests/nginx_test.go
```diff
@@ -455,17 +455,11 @@ func TestIngressControllerResources(t *testing.T) {
...
-					if version == nil {
-						return tc.Name + "-unspecified"
-					}
-					return tc.Name + version.name
-				}(),
-				func(t *testing.T) {
-					t.Parallel()
-
+		tc := tc
+		t.Run(
```
OliverMKing, line 459: this was a race condition detected by race condition unit test flag

OliverMKing: I do not know why this wasn't discovered by our github pipeline because that also uses the race flag.
https://github.com/Azure/aks-app-routing-operator/blob/5109e8158b8136eb1250134516e537f025c343c1/.github/workflows/unit.yaml#L51

jaiveerk (approved):
from what I understand this is slightly different from what asm does (which leverages env vars) but if this has the same effect then lgtm

OliverMKing:
> from what I understand this is slightly different from what asm does (which leverages env vars) but if this has the same effect then lgtm

It's the same as the asm approach. This injects the env variables. We can't follow the asm approach (which is namespace based) because we use CRDs and customers can choose the namespace.

jaiveerk:
> > from what I understand this is slightly different from what asm does (which leverages env vars) but if this has the same effect then lgtm
> 
> It's the same as the asm approach. This injects the env variables. We can't follow the asm approach (which is namespace based) because we use CRDs and customers can choose the namespace.

sg, thanks for the callout with the docs

PR #437, opened 2025-05-27
add load balancer source ip ranges feature

OliverMKing:
# Description

adds the load balancer source ip ranges feature

### DIFF
--- a/pkg/manifests/nginx.go
+++ b/pkg/manifests/nginx.go
@@ -307,10 +307,13 @@ func newNginxIngressControllerRoleBinding(conf *config.Config, ingressConfig *Ng
 
 func newNginxIngressControllerService(conf *config.Config, ingressConfig *NginxIngressConfig) *corev1.Service {
 	annotations := make(map[string]string)
+	sourceRanges := []string{}
 	if ingressConfig != nil && ingressConfig.ServiceConfig != nil {
 		for k, v := range ingressConfig.ServiceConfig.Annotations {
 			annotations[k] = v
 		}
+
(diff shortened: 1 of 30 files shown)


pkg/manifests/nginx_test.go
```diff
@@ -40,6 +40,7 @@ var (
 			Annotations: map[string]string{
 				"service.beta.kubernetes.io/azure-load-balancer-internal": "true",
 			},
+			LoadBalancerSourceRanges: []string{"127.1.000.1"},
```
jaiveerk, line 43: can we add a case where this isn't specified to ensure we're testing default behavior?

jaiveerk (changes requested):

pkg/manifests/nginx.go
```diff
@@ -307,10 +307,13 @@ func newNginxIngressControllerRoleBinding(conf *config.Config, ingressConfig *Ng
 
 func newNginxIngressControllerService(conf *config.Config, ingressConfig *NginxIngressConfig) *corev1.Service {
 	annotations := make(map[string]string)
+	sourceRanges := []string{}
 	if ingressConfig != nil && ingressConfig.ServiceConfig != nil {
 		for k, v := range ingressConfig.ServiceConfig.Annotations {
 			annotations[k] = v
 		}
+
+		sourceRanges = ingressConfig.ServiceConfig.LoadBalancerSourceRanges
```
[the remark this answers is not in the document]
jaiveerk: whoops lol sorry I read over it too fast

pkg/manifests/nginx_test.go
OliverMKing: done

PR #438, opened 2025-05-28
cleanup managed cert feature

OliverMKing:
# Description

cleans up unneeded code that was planned on being used for managed cert feature and won't be used for fully autogenerated domain pivot.


jaiveerk (dismissed):
lgtm

PR #440, opened 2025-05-30
add ssl passthrough configuration option

OliverMKing:
# Description

adds ssl passthrough NginxIngressController configuration option

### DIFF
--- a/pkg/manifests/nginx.go
+++ b/pkg/manifests/nginx.go
@@ -428,7 +428,7 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 	}
 
 	if ingressConfig.EnableSSLPassthrough {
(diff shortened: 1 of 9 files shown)


pkg/manifests/nginx.go
```diff
@@ -427,6 +427,10 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 		deploymentArgs = append(deploymentArgs, "--default-backend-service="+ingressConfig.DefaultBackendService)
 	}
 
+	if ingressConfig.EnableSSLPassthrough {
```
jaiveerk, line 430: optional bool defaults to false, right?

OliverMKing: yes

jaiveerk (approved):
lgtm nice work

PR #451, opened 2025-06-23
enable oidc in e2e clusters

OliverMKing:
# Description

enables oidc in e2e clusters so we can test workload identity


jaiveerk (approved):
i love small prs!

PR #469, opened 2025-07-25
add toggle for default domain certificate install conditionally

OliverMKing:
# Description

add toggle for default domain certificate


meecethereese (approved):
lgtm

PR #471, opened 2025-07-29
add store that loads local files into memory and periodically refreshes them

OliverMKing:
# Description

adds a store that loads local files into memory and periodically refreshes them. This will be used to load the default domain certificate and handle any rotations when they happen.


pkg/config/config.go
```diff
@@ -110,6 +111,16 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("crd path %s is not a directory", c.CrdPath)
 	}
 
+	if !c.EnableDefaultDomain && c.DefaultDomainCertPath != "" {
+		return errors.New("--default-domain-cert-path is not allowed when --enable-default-domain is not set")
+	}
+
+	if c.EnableDefaultDomain {
+		if c.DefaultDomainCertPath == "" {
```
jaiveerk, line 119: style nit - why not && these two?

pkg/config/config_test.go
```diff
@@ -274,6 +274,84 @@ var validateTestCases = []struct {
...
+			MSIClientID:              "test-msi-client-id",
+			TenantID:                 "test-tenant-id",
+			Cloud:                    "test-cloud",
+			Location:                 "test-location",
+			ConcurrencyWatchdogThres: 101,
+			ConcurrencyWatchdogVotes: 2,
+			ClusterUid:               "cluster-uid",
+			OperatorDeployment:       "app-routing-operator",
+			CrdPath:                  validCrdPath,
+			EnableDefaultDomain:      false,
```
jaiveerk, line 351: since enableddefaultdomainname and defaultdomaincertpath are default values this case is redundant right? already covered by the valid full and valid minimal cases?

pkg/controller/controller.go
```diff
@@ -147,6 +154,18 @@ func NewManagerForRestConfig(conf *config.Config, rc *rest.Config) (ctrl.Manager
 		return nil, fmt.Errorf("setting up indexers: %w", err)
 	}
 
+	storeLog := m.GetLogger().WithName("store")
+	store := store.New(storeLog, context.Background(), storeRefreshInterval)
```
jaiveerk, line 158: what's the case for enabling store when enabledefaultdomain isn't also enabled?

pkg/store/store.go
```diff
@@ -0,0 +1,168 @@
...
+	return s
+}
+
+// AddFile adds a local file to the store for tracking
+func (s *store) AddFile(path string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// Check if file exists
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
jaiveerk, line 55: we should catch non-isnotexist errors too

pkg/store/store.go
```diff
@@ -0,0 +1,168 @@
...
+		s.logger.Info("Removed file from store", "path", path)
+	}
+}
+
+// GetContent returns just the content bytes for the given path
+func (s *store) GetContent(path string) ([]byte, bool) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	file, exists := s.files[path]
```
jaiveerk, line 90: another style nit - feel like a named return could be useful here

pkg/store/store.go
```diff
@@ -0,0 +1,168 @@
...
+	content, err := os.ReadFile(file.Path)
+	if err != nil {
+		return fmt.Errorf("failed to read updated file %s: %w", file.Path, err)
+	}
+
+	oldSize := len(file.Content)
+	file.Content = content
+
+	s.logger.Info("Refreshed file content",
+		"path", path,
```
jaiveerk, line 122: is this adding properties to the logger? If so I feel like we should just include this in the msg itself vs adding more columns

pkg/store/store_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	err := os.WriteFile(testFile, []byte("initial"), 0o644)
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	store := New(logr.Discard(), ctx, 0) // 0 interval disables periodic refresh
+	err = store.AddFile(testFile)
+	require.NoError(t, err)
+
+	// Test concurrent read/write access
```
jaiveerk, line 135: how are we testing write logic here? wouldn't we want to have a "write" event after which we expect both readers are consistent?

jaiveerk (changes requested):
small stuff but overall design is solid

pkg/config/config_test.go
OliverMKing: yup

pkg/store/store_test.go
OliverMKing: updated to use fssync

pkg/store/store.go
```diff
@@ -0,0 +1,247 @@
...
+			// Send error to error channel
+			select {
+			case s.errorCh <- err:
+			default:
+				s.logger.Info("Error channel full, dropping error")
+			}
+		} else {
+			// Send rotation event for file update
+			select {
+			case s.rotationCh <- RotationEvent{Path: path}:
```
jaiveerk, line 213: where do we leverage rotation events? or will they eventually be emitted to other resources?

pkg/store/store.go
```diff
@@ -0,0 +1,247 @@
...
+	if !exists {
+		return
+	}
+
+	s.logger.Info("File event received", "path", path, "op", event.Op.String())
+
+	// Handle different event types
+	if event.Has(fsnotify.Remove) || event.Has(fsnotify.Rename) {
+		// File was deleted or renamed - return error
+		err := fmt.Errorf("file was deleted or renamed: %s", path)
```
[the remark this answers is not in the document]
OliverMKing: No. If the file name changes it's a different file basically. Something went very wrong

pkg/config/config_test.go
OliverMKing: Explicit test case is fine

pkg/store/store.go
OliverMKing: They'll be used to requeue managed domain controller. Will be needed

OliverMKing: Error one is for health check to ensure everything is healthy

pkg/store/store_test.go
```diff
@@ -0,0 +1,399 @@
...
+	// Ensure no rotation events occurred for deletion
+	select {
+	case event := <-rotationEvents:
+		t.Fatalf("Unexpected rotation event for file deletion: %+v", event)
+	default:
+		// No rotation event, as expected
+	}
+}
+
+func TestStore_FileDeletedExternally(t *testing.T) {
```
meecethereese, line 327: I think you could get rid of this unit test. Most of it is covered in TestStore_FileDeletedExternallyGeneratesError() and what isn't covered there (checking for no rotation event) gets covered in TestStore_RotationEvents(). Or maybe get rid of TestStore_FileDeletedExternallyGeneratesError()?

pkg/store/store.go
```diff
@@ -0,0 +1,251 @@
...
```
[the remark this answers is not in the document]
OliverMKing: 182 is checking if it's in the store and if we should be tracking this file. This block is hit like the code says

jaiveerk (approved):

PR #474, opened 2025-07-31
add tls package

OliverMKing:
# Description

add tls package. This will be used to validate that the certs are valid before reconciling.

Will also be used by E2E to validate


pkg/tls/tls_test.go
```diff
@@ -0,0 +1,136 @@
...
+	return certPEM, keyPEM
+}
+
+func TestParseTLSCertificate_Valid(t *testing.T) {
+	now := time.Now()
+	notBefore := now.Add(-24 * time.Hour)
+	notAfter := now.Add(24 * time.Hour)
+	dnsNames := []string{"example.com", "www.example.com"}
+
+	certPEM, keyPEM := generateTestCertificate(t, notBefore, notAfter, dnsNames)
```
jaiveerk, line 73: maybe a good idea to feed template into this function and then compare the parsed certificate from template which serves as the ground truth? 

Or better yet - we marshal a known/valid cert struct to bytes then use that initial cert struct as the ground truth and compare parsed result to that

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,136 @@
...
+		t.Error("Expected certificate to be expired")
+	}
+}
+
+func TestParseTLSCertificate_InvalidCertPEM(t *testing.T) {
+	invalidCertPEM := []byte("invalid certificate")
+	_, keyPEM := generateTestCertificate(t, time.Now().Add(-24*time.Hour), time.Now().Add(24*time.Hour), []string{"example.com"})
+
+	_, err := ParseTLSCertificate(invalidCertPEM, keyPEM)
+	if err == nil {
```
jaiveerk, line 123: nit: specific error checking

jaiveerk (changes requested):

pkg/tls/tls_test.go
jaiveerk: not hardcoded but I'm saying that the "actual" value (output from ParseTLSCertificate) that we compare should come from an explicitly defined expectation. 

Like if we define our expectation with x, why use f(x) for comparison? If f works as intended then x should contain all the information that we eventually want to use for comparison anyway

jaiveerk (approved):

PR #479, opened 2025-08-06
add requeue on rotation to default domain cert controller

OliverMKing:
# Description

adds requeue on rotation to default domain cert controller. Mauricio is working on e2e test for this


meecethereese (approved):
lgtm

PR #481, opened 2025-08-08
fix k8s secret rotations, k8s secret volume mount rotations are treated as deletes then recreates

OliverMKing:
# Description

fix k8s secret rotations, k8s secret volume mount rotations are treated as deletes then recreates

### DIFF
--- a/pkg/store/store_test.go
+++ b/pkg/store/store_test.go
@@ -397,3 +397,191 @@ func TestStore_ChannelAccessors(t *testing.T) {
 		// Expected behavior - no events yet
 	}
 }
+
+func TestStore_KubernetesSecretRotation(t *testing.T) {
+	// Create a temporary directory to simulate K8s secret mount
+	tmpDir := t.TempDir()
+
+	// Create initial secret file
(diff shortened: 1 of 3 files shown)


pkg/store/store_test.go
```diff
@@ -397,3 +397,191 @@ func TestStore_ChannelAccessors(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = store.AddFile(secretFile)
+	require.NoError(t, err)
+
+	// Remove the file
+	err = os.Remove(secretFile)
+	require.NoError(t, err)
+
+	// Create new file but in a way that might cause watcher issues
```
meecethereese, line 567: Is the edge case here just about handling a specific error?

meecethereese (approved):
lgtm

pkg/store/store_test.go
OliverMKing: I just decided to remove it. AI added it and I don't think I really liked it. Was non deterministic

meecethereese (approved):

PR #482, opened 2025-08-11
simulate atomic write file for store

OliverMKing:
# Description

behave exactly like atomic write file for a more complete unit test for the store to handle k8s secret rotations


meecethereese (approved):
lgtm

PR #577, opened 2026-04-07
Add dalec image of ingress nginx

meecethereese:
# Description

This PR adds a flag to enable the use of the Dalec image of ingress-nginx as well as incorporating the underlying logic to make that flag functional.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -41,6 +41,7 @@ func init() {
 	flag.IntVar(&Flags.ConcurrencyWatchdogVotes, "concurrency-watchdog-votes", 4, "number of votes required for a pod to be considered for load shedding")
 	flag.BoolVar(&Flags.DisableOSM, "disable-osm", false, "enable Open Service Mesh integration")
 	flag.BoolVar(&Flags.DisableIngressNginx, "disable-ingress-nginx", false, "disable the ingress-nginx integration")
+	flag.BoolVar(&Flags.EnableDalecNginx, "enable-dalec-nginx", false, "use dalec-built nginx ingress controller image")
 	flag.StringVar(&Flags.ServiceAccountTokenPath, "service-account-token-path", "", "optionally override the default token path")
 	flag.StringVar(&Flags.MetricsAddr, "metrics-addr", "0.0.0.0:8081", "address to serve Prometheus metrics on")
 	flag.StringVar(&Flags.ProbeAddr, "probe-addr", "0.0.0.0:8080", "address to serve readiness/liveness probes on")
--- a/pkg/manifests/nginx.go
+++ b/pkg/manifests/nginx.go
@@ -27,6 +27,9 @@ var (
 	}
 	nginxVersionsAscending = []NginxIngressVersion{nginx1_13_7}
 	LatestNginxVersion     = nginxVersionsAscending[len(nginxVersionsAscending)-1]
+
+	nginxImagePath      = "/oss/kubernetes/ingress/nginx-ingress-controller:"
+	dalecNginxImagePath = "/oss/v2/ingress-nginx/controller:"
 )
 
 var nginxLabels = util.MergeMaps(
@@ -381,6 +384,21 @@ func newNginxIngressControllerPromService(conf *config.Config, ingressConfig *Ng
 	}
 }
 
+func nginxImage(conf *config.Config, ingressConfig *NginxIngressConfig) string {
+	imagePath := nginxImagePath
+	if conf.EnableDalecNginx {
+		imagePath = dalecNginxImagePath
+	}
+	return path.Join(conf.Registry, imagePath+ingressConfig.Version.tag)
+}
+
+func nginxRunAsUser(conf *config.Config) *int64 {
+	if conf.EnableDalecNginx {
+		return util.Int64Ptr(1000)
+	}
+	return util.Int64Ptr(101)
+}
+
 func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *NginxIngressConfig) *appsv1.Deployment {
 	ingressControllerDeploymentLabels := AddComponentLabel(GetTopLevelLabels(), IngressControllerComponentName)
 
@@ -467,7 +485,7 @@ func newNginxIngressControllerDeployment(conf *config.Config, ingressConfig *Ngi
 					ServiceAccountName: ingressConfig.ResourceName,
 					Containers: []corev1.Container{*withPodRefEnvVars(withLivenessProbeMatchingReadinessNewFailureThresh(withTypicalReadinessProbe(10254, &corev1.Container{
 						Name:  "controller",
-						Image: path.Join(conf.Registry, "/oss/kubernetes/ingress/nginx-ingress-controller:"+ingressConfig.Version.tag),
+						Image: nginxImage(conf, ingressConfig),
 						Args:  deploymentArgs,
 						SecurityContext: &corev1.SecurityContext{
 							AllowPrivilegeEscalation: util.ToPtr(false),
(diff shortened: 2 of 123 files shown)


OliverMKing (commented):
this LGTM. let's make e2es run on dalec image.

One thing -- we might need to decouple the image version. on dalec we're jumping to something different than normal

pkg/manifests/nginx.go
```diff
@@ -61,7 +71,11 @@ const internalLogFormat = `{"remote_addr":"$remote_addr","remote_user":"$remote_
 
 func GetNginxResources(conf *config.Config, ingressConfig *NginxIngressConfig) *NginxResources {
 	if ingressConfig != nil && ingressConfig.Version == nil {
-		ingressConfig.Version = &LatestNginxVersion
+		if conf.EnableDalecNginx {
+			ingressConfig.Version = &LatestDalecNginxVersion
```
OliverMKing: **Bug: Version/image mismatch when explicit version is set with dalec flag enabled**

`nginxImage()` and `nginxRunAsUser()` are always gated on `conf.EnableDalecNginx`, but the dalec flag only affects version selection when `ingressConfig.Version == nil`. If someone explicitly sets `Version = &nginx1_13_7` (the standard version) while `EnableDalecNginx = true`, the resulting image path will be `/oss/v2/ingress-nginx/controller:v1.13.7` — a dalec image path with a non-dalec version tag that likely doesn't exist.

**Suggestion:** Either validate that an explicitly-set version belongs to `dalecNginxVersionsAscending` when the dalec flag is on, or override the version entirely when the flag is enabled (not just when it's nil).

pkg/config/config.go
```diff
@@ -41,6 +41,7 @@ func init() {
 	flag.IntVar(&Flags.ConcurrencyWatchdogVotes, "concurrency-watchdog-votes", 4, "number of votes required for a pod to be considered for load shedding")
 	flag.BoolVar(&Flags.DisableOSM, "disable-osm", false, "enable Open Service Mesh integration")
 	flag.BoolVar(&Flags.DisableIngressNginx, "disable-ingress-nginx", false, "disable the ingress-nginx integration")
+	flag.BoolVar(&Flags.EnableDalecNginx, "enable-dalec-nginx", false, "use dalec-built nginx ingress controller image")
```
OliverMKing, line 44: **No e2e coverage for the dalec image path**

The e2e infrastructure is wired up (`withDalecNginx()` builder in `testing/e2e/suites/operatorConfig.go`, `--enable-dalec-nginx` arg in `testing/e2e/manifests/operator.go`), but no e2e test suite actually calls `withDalecNginx(true)`. This means the dalec image is never exercised in e2e — there's no integration coverage for the new image actually starting and serving traffic.

**Suggestion:** Add at least one e2e suite that enables the dalec flag, even if gated behind an environment variable or build tag.

pkg/manifests/nginx.go
OliverMKing: i don't really agree with this comment. seems fine as is to me, closing

OliverMKing: I do think the abstraction could be better. we could have a dalec specific version struct and the functions could be methods on that and version could be an interface but this is non blocking

pkg/config/config.go
meecethereese: Ended up resolving this not by adding another e2e suite but by making the operator configs running on the latest version use the dalec image. Each e2e suite has multiple configs so now each one is testing both images at different versions

pkg/manifests/nginx.go
```diff
@@ -61,7 +71,14 @@ const internalLogFormat = `{"remote_addr":"$remote_addr","remote_user":"$remote_
 
 func GetNginxResources(conf *config.Config, ingressConfig *NginxIngressConfig) *NginxResources {
 	if ingressConfig != nil && ingressConfig.Version == nil {
-		ingressConfig.Version = &LatestNginxVersion
+		// copy to avoid mutating the caller's struct (prevents data races)
+		cfgCopy := *ingressConfig
+		if conf.EnableDalecNginx {
```
jaiveerk, line 76: why do we have this decision here instead of within the deployment? that should have access to the top level config struct too

jaiveerk (changes requested):

pkg/manifests/nginx.go
jaiveerk: like this seems redundant given that we have the function for nginxImage anyway

testing/e2e/infra/infras.go
```diff
@@ -44,6 +44,7 @@ var Infras = infras{
 		FederatedNamespaces: GenerateGatewayFederatedNamespaces(),
 		PostCreate:          clients.EnableAppRoutingIstio,
 	},
+
```
jaiveerk: nit remove this

testing/e2e/manifests/operator.go
```diff
@@ -157,6 +157,12 @@ func (o OperatorVersion) String() string {
 	}
 }
 
+// SupportsDalec reports whether this operator version supports the --enable-dalec-nginx flag.
+// Update the threshold here when a new released version ships with dalec support.
+func (o OperatorVersion) SupportsDalec() bool {
+	return o >= OperatorVersionLatest
```
jaiveerk, line 163: can we preemptively add an OR case to future proof this where we say it's greater than or equal to latest OR it's greater than or equal to the next version?

testing/e2e/manifests/operator.go
```diff
@@ -235,6 +243,7 @@ func (o *OperatorConfig) args(publicZones, privateZones []string) []string {
 
 	if o.Version >= OperatorVersionLatest {
 		enableGatewayArg = "--enable-gateway-tls"
+		enableDalecNginxArg = "--enable-dalec-nginx"
```
jaiveerk: sounds like we should be using supportsDalec instead to be able to determine whether or not we want this field?

testing/e2e/manifests/operator.go
meecethereese: I like that idea, but given that currently we're only supporting the latest version I'm not sure what I would put for the "next version" to complete the OR

testing/e2e/manifests/operator.go
meecethereese: Yeah that's a good callout. It'll future proof that arg too once we're in a position to update SupportsDalec

pkg/manifests/nginx.go
meecethereese: I think my original reasoning for positioning the logic to set the Version struct here was that it's the same place it used to be. I could move this logic into the nginxImage() function but if you take a look at [line 99](https://github.com/Azure/aks-app-routing-operator/blob/c9924c3278cb5d650a558a183aa49301ce4e12b6/pkg/manifests/nginx.go#L82), it does seem to have more merit to keep it here as it would then be needed there.

To be honest, I'm not sure what you meant when you said "within the deployment"

jaiveerk: if you perform this logic where you read conf.EnableDalecNginx within the function called "newNginxIngressControllerDeployment" we have no reason to copy the config struct or overwrite the version field.

meecethereese: Honestly, I think you're right. However, I think what you're pointing out is actually a bigger issue with the  ingressConfig struct and would be best addressed in a separate PR. If I choose to not set Version at all and just rely on the LatestDalecNginxVersion and LatestNginxVersion structs to create the image path, it's exposing the reality that Version is a useless field in the ingressConfig struct. I think the only reason it exists is for that switch case on line 99, but that has no logic in it. Once I remove its use, I'd also want to remove it from the tests and everything too.

So, since at the moment these if-elses aren't exact copies and the logic works as is I'll leave it this way and put down the ingressConfig Version rework as a future code improvement item.

OliverMKing (approved):

pkg/manifests/nginx.go
```diff
@@ -381,6 +398,20 @@ func newNginxIngressControllerPromService(conf *config.Config, ingressConfig *Ng
 	}
 }
 
+func nginxImage(conf *config.Config, ingressConfig *NginxIngressConfig) string {
```
jaiveerk, line 401: can we have UTs for this to ensure coverage doesn't drop too much

jaiveerk (approved):

PR #595, opened 2026-05-13
Adding E2e Tests for GrpcRoute Type

jaiveerk:
# Description

Adding dns/tls e2e scenario for grpc routes as well as httproutes, to ensure externaldns works correctly for resolving both kinds of routes.

Fixes # (issue)
Feature # (details)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,4 @@ job-app-routing-operator-e2e.log
 *.DS_Store
 
 # Claude
-./claude/scheduled_tasks.lock
+./claude/scheduled_tasks.lock.claude/
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,4 @@ job-app-routing-operator-e2e.log
 *.DS_Store
 
 # Claude
-./claude/scheduled_tasks.lock.claude/
+./claude/scheduled_tasks.lock
--- a/testing/e2e/manifests/embedded/grpc_gateway_client.golang
+++ b/testing/e2e/manifests/embedded/grpc_gateway_client.golang
@@ -109,13 +109,12 @@ func main() {
 				w.WriteHeader(500)
 				return
 			}
-			// Prefer typed checks over substring matches on err.Error() so we do not silently
-			// accept unexpected errors if the stdlib ever rewords its messages. DNS NXDOMAIN
-			// surfaces as *net.DNSError; if the resolver retries past our 10s budget, we get
-			// context.DeadlineExceeded.
-			var dnsErr *net.DNSError
-			if !errors.As(err, &dnsErr) && !errors.Is(err, context.DeadlineExceeded) {
-				log.Printf("ERROR: expected no such host or deadline exceeded for %s, got: %v", unreachableHost, err)
+			// grpc-go wraps the underlying *net.DNSError in a *status.Error with
+			// codes.Unavailable, and errors.As does NOT unwrap through it. Check the grpc
+			// status code instead, plus context.DeadlineExceeded as a fallback if the
+			// resolver retries past our 10s budget.
+			if status.Code(err) != codes.Unavailable && !errors.Is(err, context.DeadlineExceeded) {
+				log.Printf("ERROR: expected Unavailable or deadline exceeded for %s, got: %v", unreachableHost, err)
 				w.WriteHeader(500)
 				return
 			}
--- a/testing/e2e/manifests/embedded/grpc_gateway_client.golang
+++ b/testing/e2e/manifests/embedded/grpc_gateway_client.golang
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"log"
 	"net"
 	"net/http"
@@ -52,10 +53,13 @@ func loadRoots() *x509.CertPool {
 func dialAndCheck(ctx context.Context, host string, dialer func(ctx context.Context, addr string) (net.Conn, error), roots *x509.CertPool) error {
 	creds := credentials.NewTLS(&tls.Config{ServerName: host, RootCAs: roots})
 	target := host + ":443"
-	conn, err := grpc.DialContext(ctx, target,
+	// grpc.NewClient is the modern replacement for grpc.DialContext. It does not block on the
+	// initial connection — the runtime connects lazily on the first RPC, which means a DNS
+	// failure surfaces cleanly from Health/Check rather than getting masked by a WithBlock
+	// "context deadline exceeded".
+	conn, err := grpc.NewClient(target,
 		grpc.WithTransportCredentials(creds),
 		grpc.WithContextDialer(dialer),
-		grpc.WithBlock(),
 	)
 	if err != nil {
 		return err
@@ -104,10 +108,12 @@ func main() {
 				w.WriteHeader(500)
 				return
 			}
-			// Accept either "no such host" (DNS lookup failed) or "context deadline exceeded"
-			// (grpc.WithBlock retries the dial past DNS NXDOMAIN until the deadline). Both
-			// indicate the host is correctly unreachable.
-			if !strings.Contains(err.Error(), "no such host") && !strings.Contains(err.Error(), "context deadline exceeded") {
+			// Prefer typed checks over substring matches on err.Error() so we don't silently
+			// accept unexpected errors if the stdlib ever rewords its messages. DNS NXDOMAIN
+			// surfaces as *net.DNSError; if the resolver retries past our 10s budget, we get
+			// context.DeadlineExceeded.
+			var dnsErr *net.DNSError
+			if !errors.As(err, &dnsErr) && !errors.Is(err, context.DeadlineExceeded) {
 				log.Printf("ERROR: expected no such host or deadline exceeded for %s, got: %v", unreachableHost, err)
 				w.WriteHeader(500)
 				return
--- a/testing/e2e/manifests/embedded/grpc_gateway_client.golang
+++ b/testing/e2e/manifests/embedded/grpc_gateway_client.golang
@@ -108,7 +108,7 @@ func main() {
 				w.WriteHeader(500)
 				return
 			}
-			// Prefer typed checks over substring matches on err.Error() so we don't silently
+			// Prefer typed checks over substring matches on err.Error() so we do not silently
 			// accept unexpected errors if the stdlib ever rewords its messages. DNS NXDOMAIN
 			// surfaces as *net.DNSError; if the resolver retries past our 10s budget, we get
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -16,4 +16,4 @@ job-app-routing-operator-e2e.log
 *.DS_Store
 
 # Claude
-./claude/scheduled_tasks.lock
```
OliverMKing, line 19: this doesn't look right. did you mean to add .claude? We don't necessarily want to ignore all of .claude

testing/e2e/suites/gateway.go
```diff
@@ -57,6 +57,49 @@ type multiZoneGatewayTestConfig struct {
...
+// recordPrefix is the hostname prefix used to derive DNS record names. Shared across route
+// kinds since kinds run serially within a single top-level entry (cleanup waits for record
+// deletion before the next kind starts).
+func (c multiZoneGatewayTestConfig) recordPrefix() string {
+	return "zone"
+}
+
+// buildGatewayResources builds the kind-appropriate gateway+route+client+server resource set.
+func (c multiZoneGatewayTestConfig) buildGatewayResources(namespace, name, nameserver, kvURI, tlsHost, sa, gwClass string) manifests.GatewayClientServerResources {
+	if c.routeKind != nil && c.routeKind.Name() == "grpc" {
```
OliverMKing: The dispatch here defeats the `RouteKind` abstraction this PR introduces. `GatewayGrpcClientAndServer` is a one-liner wrapper around `GatewayClientAndServerFor(GRPCRouteKind{}, ...)`, and the kind interface already exposes `ClientContents()`/`ServerContents()`/`ServicePortName()`. Suggest replacing both branches with:

```go
return manifests.GatewayClientAndServerFor(c.routeKind, manifests.GatewayClientServerArgs{...})
```

Then `GatewayGrpcClientAndServer` and the name-based switch can be deleted. As written, adding TLSRoute requires another wrapper + another string match.

testing/e2e/manifests/gatewayClientServer.go
```diff
@@ -353,9 +373,11 @@ func buildGatewayClient(namespace, name, url, nameserver, unreachableURL, tlsSec
 	deployment.Spec.Template.Spec.Containers[0].ReadinessProbe = &corev1.Probe{
 		FailureThreshold:    1,
 		InitialDelaySeconds: 1,
-		PeriodSeconds:       1,
+		PeriodSeconds:       5,
 		SuccessThreshold:    1,
-		TimeoutSeconds:      5,
+		// 30s gives the gRPC client (which dials TLS with WithBlock and may also test an
+		// unreachable host) enough room without the kubelet canceling the request mid-flight.
```
OliverMKing: This widens probe timing (`PeriodSeconds` 1→5, `TimeoutSeconds` 5→30) for **all** gateway clients, not just gRPC. HTTPRoute tests previously got 1s feedback and will now poll every 5s with a 30s ceiling, slowing the HTTP suite unnecessarily. Consider parameterizing per kind — e.g. take probe settings from the `RouteKind` (a `ProbeTuning()` method), or only widen when `kind.Name() == "grpc"`.

testing/e2e/manifests/embedded/grpc_gateway_client.golang
```diff
@@ -0,0 +1,120 @@
...
+
+		if unreachableHost != "" {
+			ctx2, cancel2 := context.WithTimeout(r.Context(), 10*time.Second)
+			defer cancel2()
+			err := dialAndCheck(ctx2, unreachableHost, dialer, roots)
+			if err == nil {
+				log.Printf("ERROR: expected unreachable %s to fail, but it succeeded", unreachableHost)
+				w.WriteHeader(500)
+				return
+			}
```
OliverMKing, line 113: `grpc.DialContext` and `grpc.WithBlock` are deprecated in current grpc-go — `grpc.NewClient` is the replacement, and blocking on dial is no longer recommended (the runtime handles reconnects lazily). Not a blocker for e2e, but you'll see deprecation warnings on the next grpc-go bump, and `WithBlock` is what forces the awkward "context deadline exceeded" fallback in the unreachable-host check below — without it, a clean DNS failure would surface immediately.

testing/e2e/manifests/embedded/grpc_gateway_client.golang
```diff
@@ -0,0 +1,120 @@
...
+			// Accept either "no such host" (DNS lookup failed) or "context deadline exceeded"
+			// (grpc.WithBlock retries the dial past DNS NXDOMAIN until the deadline). Both
+			// indicate the host is correctly unreachable.
+			if !strings.Contains(err.Error(), "no such host") && !strings.Contains(err.Error(), "context deadline exceeded") {
```
OliverMKing: Matching on `err.Error()` substrings couples the test to net package phrasing — if the Go stdlib ever changes the wording ("no such host" → "host not found", etc.) this will silently start accepting unexpected errors. Prefer typed checks:

```go
var dnsErr *net.DNSError
if !errors.As(err, &dnsErr) && !errors.Is(err, context.DeadlineExceeded) {
    // unexpected failure
}
```

`net.DNSError.IsNotFound` can further narrow to NXDOMAIN if desired.

testing/e2e/manifests/route_kind.go
```diff
@@ -7,6 +7,70 @@ import (
...
+func (GRPCRouteKind) ServerContents() string  { return grpcServerContents }
+func (GRPCRouteKind) ServicePortName() string { return "grpc" }
+
+func (GRPCRouteKind) RouteObjectName(baseName string) string {
+	return baseName + "-grpcroute"
+}
+
+func (GRPCRouteKind) Listener(listenerName, tlsHost, keyvaultURI, serviceAccountName string) gatewayv1.Listener {
+	// Identical to HTTPRouteKind — both kinds ride HTTPS/443 with TLS terminate.
+	return HTTPRouteKind{}.Listener(listenerName, tlsHost, keyvaultURI, serviceAccountName)
```
OliverMKing: Delegating to `HTTPRouteKind{}.Listener` couples gRPC behavior to whatever HTTPRoute decides in the future — e.g. if HTTPRoute later adds HTTP/1.1-specific ALPN or filter config, gRPC will silently inherit it. Suggest extracting the shared bit into a package-level helper (e.g. `httpsTerminateListener(...)`) that both kinds call explicitly, so changes to one kind don't leak into the other.

testing/e2e/manifests/route_kind.go
```diff
@@ -7,6 +7,70 @@ import (
...
+var _ RouteKind = GRPCRouteKind{}
+
+func (GRPCRouteKind) Name() string { return "grpc" }
+
+func (GRPCRouteKind) ClientContents() string  { return grpcGatewayClientContents }
+func (GRPCRouteKind) ServerContents() string  { return grpcServerContents }
+func (GRPCRouteKind) ServicePortName() string { return "grpc" }
+
+func (GRPCRouteKind) RouteObjectName(baseName string) string {
+	return baseName + "-grpcroute"
```
OliverMKing, line 24: `RouteObjectName` appends `-grpcroute` (10 chars) vs HTTPRoute's `-httproute`. Combined with `zoneTypePrefix + "zone" + zoneIndex` upstream (`gateway.go` ~line 669), per-kind resource names can get long. Worth sanity-checking the worst-case length against the 63-char DNS-label limit for any derived names (Services, pods). Since kinds run serially in the same namespace, a collision would manifest as a confusing "already exists" mid-run rather than a clean failure.

.gitignore
jaiveerk: yeah I wasn't sure what was up with this either -- claude made the change and thought it was due to a rename. but will revert thanks for the calloutyeah I wasn't sure what was up with this either -- claude made the change and thought it was due to a rename. but will revert thanks for the callout

testing/e2e/manifests/route_kind.go
jaiveerk: yeah good call will update

testing/e2e/manifests/route_kind.go
jaiveerk: the test already ran successfully so I don't think this is an issue

testing/e2e/manifests/gatewayClientServer.go
jaiveerk: yeah good call here

testing/e2e/manifests/embedded/grpc_gateway_client.golang
jaiveerk: great call

testing/e2e/manifests/embedded/grpc_gateway_client.golang
jaiveerk: good call, updating

testing/e2e/suites/gateway.go
jaiveerk: great call yeah will do

testing/e2e/manifests/embedded/grpc_gateway_client.golang
```diff
@@ -0,0 +1,126 @@
...
+			return d.DialContext(ctx, "tcp", ns+":53")
+		},
+	}
+	d := &net.Dialer{Resolver: resolver, Timeout: 10 * time.Second}
+	return func(ctx context.Context, addr string) (net.Conn, error) {
+		return d.DialContext(ctx, "tcp", addr)
+	}
+}
+
+func loadRoots() *x509.CertPool {
```
OliverMKing: `loadRoots` silently falls back to system roots when the mounted cert can't be read or parsed (returning nil → `tls.Config.RootCAs == nil`). Since this client specifically exists to assert the mounted KV cert is valid, a real cert-mount bug could turn into a passing probe. Suggest returning an error here and failing the probe instead of warning + continuing.

testing/e2e/manifests/embedded/grpc_gateway_client.golang
```diff
@@ -0,0 +1,126 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func makeDialer(nameserver string) func(ctx context.Context, addr string) (net.Conn, error) {
+	resolver := &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			d := net.Dialer{Timeout: time.Second}
+			ns := nameserver
+			if strings.HasSuffix(ns, ".") {
```
OliverMKing: Nit: `ns = strings.TrimSuffix(ns, ".")` is equivalent and more idiomatic than the HasSuffix + slice.

OliverMKing (approved):

PR #605, opened 2026-06-13
clean up default domain dns when off

OliverMKing:
Disabling the default domain feature deletes the Azure DNS zone and revokes the managed identity's RBAC, but left the
default-domain-dns-external-dns Deployment running, so it crash-looped indefinitely on 403 AuthorizationFailed.

When --enable-default-domain is off, delete the orphaned default-domain-dns ClusterExternalDNS (which cascades to its owned Deployment via owner references) and the default-domain-sa ServiceAccount at startup, mirroring the existing removeDisabledCRDs teardown.

### DIFF
--- a/pkg/controller/dns/default_domain_cluster_external_dns.go
+++ b/pkg/controller/dns/default_domain_cluster_external_dns.go
@@ -1,12 +1,18 @@
 package dns
 
 import (
+	"context"
+	"fmt"
+
 	approutingv1alpha1 "github.com/Azure/aks-app-routing-operator/api/v1alpha1"
 	"github.com/Azure/aks-app-routing-operator/pkg/config"
 	"github.com/Azure/aks-app-routing-operator/pkg/controller/common"
 	"github.com/Azure/aks-app-routing-operator/pkg/controller/controllername"
 	"github.com/Azure/aks-app-routing-operator/pkg/manifests"
+	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


pkg/controller/dns/default_domain_cluster_external_dns.go
```diff
@@ -21,6 +27,38 @@ const (
...
+	toDelete := []client.Object{
+		&approutingv1alpha1.ClusterExternalDNS{
+			ObjectMeta: metav1.ObjectMeta{Name: defaultDomainDNSResourceName, Namespace: conf.NS},
+		},
+		&corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{Name: defaultDomainServiceAccountName, Namespace: conf.NS},
+		},
+	}
+
+	for _, obj := range toDelete {
```
jaiveerk, line 48: do we want to delete the crd as well? or just the managed resources?

OliverMKing: out of scope for this pr. we should when someone disables app routing as a whole but that's a very different flow than this one.

OliverMKing: tilak + i are shipping a solution for this for ingress-nginx and it will be similar to the external dns crds

jaiveerk (approved):

PR #606, opened 2026-06-15
fix(ddcert): guard secret upsert and surface failures in status

OliverMKing:
## Summary

The `DefaultDomainCertificate` reconciler had two gaps that combined into a silent failure:

1. It overwrote **any** Secret at the target name, even one it didn't create.
2. On `util.Upsert` failure it only logged + emitted an event — it never touched the resource's `.status`.

When a pre-existing Secret at the target name had a different **immutable** field (e.g. `type`), `Upsert` failed on every reconcile, the `Available` condition was never set, and `kubectl wait --for=condition=Available` timed out with no signal as to why. This was observed in the wild: the RP-side cert issuance succeeded, but the in-cluster operator looped indefinitely on `Secret "cert" is invalid: type: Invalid value: "kubernetes.io/tls": field is immutable`.

## Changes

- **Ownership guard (managed-by labels):** before upserting, `Get` the target Secret; if it exists and lacks App Routing's top-level labels (`manifests.HasTopLevelLabels`), refuse to overwrite it. Emit a `ConflictingSecretExists` warning event, set `Available=False`, and requeue to re-check (we don't get watch events for Secrets we don't own).
- **Surface upsert failures:** when the upsert itself fails, also set `Available=False` with reason `ApplyingCertificateSecretFailed` (best-effort) so consumers and `kubectl wait` see the failure instead of an empty status. Event messages now include the Secret `namespace/name`.
- Extracted a small `setUnavailable` helper to keep the three `Available=False` paths consistent.

## Behavior change

Adopting a pre-existing **unlabeled** Secret at the target name is now **refused** rather than silently overwritten. If any flow relied on that adoption, this is where it changes.

## Design note

Per Kubernetes convention, the canonical ownership signal is the controller `OwnerReference`, not the `managed-by` label. This PR intentionally uses the label (reusing the existing `HasTopLevelLabels` helper) as the adoption guard. Switching to an ownerRef-based guard (or dropping `client.ForceOwnership` in `util.Upsert`) is a possible follow-up.

## Test plan

- [x] `make unit` (Docker, `go test -race ./...`) — all packages pass
- [x] New: `TestReconcile_ConflictingUnmanagedSecret_Refused` — foreign Secret untouched, status reports conflict, warning event fired, requeued
- [x] New: `TestReconcile_FailedToUpsertSecret_SetsUnavailableStatus` — apply failure sets `Available=False`
- [x] Updated: `TestReconcile_SecretAlreadyExists_UpdatesExistingSecret` — labeled (owned) Secret is still updated

### DIFF
--- a/pkg/controller/defaultdomaincert/default_domain_cert_controller_test.go
+++ b/pkg/controller/defaultdomaincert/default_domain_cert_controller_test.go
@@ -443,14 +443,11 @@ func TestReconcile_FailedToUpsertSecret_RecordsEvent(t *testing.T) {
 	assert.Contains(t, err.Error(), "failed to patch secret")
 	assert.Equal(t, ctrl.Result{}, result)
 
-	// Verify that the warning event was recorded
+	// Verify that the specific warning event was recorded: it must name the
+	// ApplyingCertificateSecretFailed reason, the target Secret, and the underlying error.
 	select {
 	case event := <-fakeRecorder.Events:
-		assert.Contains(t, event, "Warning")
-		assert.Contains(t, event, "ApplyingCertificateSecretFailed")
-		assert.Contains(t, event, "Failed to apply Secret")
-		assert.Contains(t, event, "for DefaultDomainCertificate")
-		assert.Contains(t, event, "failed to patch secret")
+		assert.Equal(t, "Warning ApplyingCertificateSecretFailed Failed to apply Secret "+testNamespace+"/"+testSecretName+" for DefaultDomainCertificate: failed to patch secret", event)
 	case <-time.After(100 * time.Millisecond):
 		t.Fatal("Expected event was not recorded within timeout")
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/defaultdomaincert/default_domain_cert_controller_test.go
```diff
@@ -448,7 +448,8 @@ func TestReconcile_FailedToUpsertSecret_RecordsEvent(t *testing.T) {
 	case event := <-fakeRecorder.Events:
 		assert.Contains(t, event, "Warning")
 		assert.Contains(t, event, "ApplyingCertificateSecretFailed")
-		assert.Contains(t, event, "Failed to apply Secret for DefaultDomainCertificate")
+		assert.Contains(t, event, "Failed to apply Secret")
+		assert.Contains(t, event, "for DefaultDomainCertificate")
```
jaiveerk: can we be more specific about the kind(s) of failures we're testing here vs just checking to see if any kind of event got emitted related to ddc

jaiveerk (commented):
small nit, lgtm otherwise

pkg/controller/defaultdomaincert/default_domain_cert_controller_test.go
OliverMKing: Claude AI: Good call — tightened both failure tests to assert the exact warning event string (reason + target Secret + underlying error) instead of loosely matching that some DDC-related event fired. See 893d21a.

jaiveerk (approved):

PR #608, opened 2026-06-18
Stabilize app-routing e2e readiness checks

OliverMKing:
## Summary

- Stabilize app-routing e2e readiness checks to reduce DNS/routing propagation flakes.
- Add targeted e2e env filters and optional cleanup skipping for preserved-infra repro/debug loops.
- Wait for the default nginx ingress controller/load balancer endpoint to become stable before running assertions.
- Relax client readiness probes and negative DNS/routing assertions to tolerate resolver/proxy-specific transient failure modes.
- Improve timeout diagnostics with deployment/pod/event summaries.
- Retry transient Azure federated identity credential creation failures (`429` / `5xx`).
- Ignore missing CRDs/resources during cleanDeploy pre-clean deletion so fresh clusters do not fail before CRDs exist.

## Validation

Fresh full-create validation, no reused infra:

```text
full-create serial validation COMPLETE result=PASS iterations=3

iteration=1 runId=20260617-2048-full-create-crdfix-1 PASS elapsedSec=1737
iteration=2 runId=20260617-2048-full-create-crdfix-2 PASS elapsedSec=2017
iteration=3 runId=20260617-2048-full-create-crdfix-3 PASS elapsedSec=1476
```

Run shape:

```text
cluster: gateway-full-mesh-cluster
mode: full create each iteration
E2E_OPERATOR_VERSION=latest
E2E_DEPLOY_STRATEGY=cleanDeploy
E2E_PUBLIC_ZONES=multiple
E2E_PRIVATE_ZONES=multiple
E2E_SKIP_CLEANUP=<unset>
E2E_TEST_FILTER=<unset>
```

Local validation:

```text
go test ./testing/e2e/clients ./testing/e2e/tests ./testing/e2e/suites ./testing/e2e/manifests ./testing/e2e/cmd ./cmd/e2e
```

Also passed earlier targeted preserved-infra validation 4 consecutive times:

```text
20260617-1534-targeted-corrected-crds PASS
20260617-1638-targeted-repeat-1 PASS
20260617-1638-targeted-repeat-2 PASS
20260617-1638-targeted-repeat-3 PASS
```

## Type of change

## Checklist

### DIFF
--- a/testing/e2e/manifests/clientServer.go
+++ b/testing/e2e/manifests/clientServer.go
@@ -65,11 +65,11 @@ func ClientAndServer(namespace, name, nameserver, keyvaultURI, host, tlsHost str
 		},
 	}
 	clientDeployment.Spec.Template.Spec.Containers[0].ReadinessProbe = &corev1.Probe{
-		FailureThreshold:    1,
+		FailureThreshold:    6,
 		InitialDelaySeconds: 1,
-		PeriodSeconds:       1,
+		PeriodSeconds:       5,
 		SuccessThreshold:    1,
-		TimeoutSeconds:      5,
+		TimeoutSeconds:      15,
 		ProbeHandler: corev1.ProbeHandler{
 			HTTPGet: &corev1.HTTPGetAction{
 				Path:   "/",
--- a/testing/e2e/manifests/e2e.go
+++ b/testing/e2e/manifests/e2e.go
@@ -1,6 +1,8 @@
 package manifests
 
 import (
+	"os"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -79,6 +81,7 @@ func E2e(image, loadableProvisionedJson string) []client.Object {
 								Name:  "app-routing-operator-e2e",
 								Image: image,
 								Args:  []string{"test", "--infra-file", "/infrastructure/infra-config.json"},
+								Env:   e2eEnv(),
 								VolumeMounts: []corev1.VolumeMount{
 									{
 										Name:      "infra-volume",
@@ -107,6 +110,26 @@ func E2e(image, loadableProvisionedJson string) []client.Object {
 	// set the group kind and version for each object
 	for _, obj := range ret {
 		setGroupKindVersion(obj)
+	}
+
+	return ret
+}
+
+func e2eEnv() []corev1.EnvVar {
+	keys := []string{
+		"E2E_TEST_FILTER",
+		"E2E_OPERATOR_VERSION",
+		"E2E_DEPLOY_STRATEGY",
+		"E2E_PUBLIC_ZONES",
+		"E2E_PRIVATE_ZONES",
+		"E2E_SKIP_CLEANUP",
+	}
+
+	ret := make([]corev1.EnvVar, 0, len(keys))
+	for _, key := range keys {
+		if val := os.Getenv(key); val != "" {
+			ret = append(ret, corev1.EnvVar{Name: key, Value: val})
+		}
 	}
 
 	return ret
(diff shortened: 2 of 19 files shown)


testing/e2e/manifests/embedded/gateway_client.golang
```diff
@@ -72,13 +67,10 @@ func main() {
...
-			if !strings.Contains(err.Error(), "no such host") {
-				log.Printf("ERROR: expected no such host error, but got: %s", err)
-				w.WriteHeader(500)
-				return
-			}
-			log.Printf("correctly failed to reach unreachable URL with no such host: %s", err)
+			// Any request failure is acceptable here. The positive URL above already proved
+			// DNS/TLS/routing for this test zone works, and the negative URL may fail as
+			// NXDOMAIN, timeout, or another resolver-specific error depending on timing.
+			log.Printf("correctly failed to reach unreachable URL: %s", err)
```
jaiveerk, line 73: I don't know about this -- I think we should keep explicitly checking for a no such host error vs some other potential client side failure if we want to keep this a true/tight negative test. Was this really a source of flakiness?

jaiveerk (changes requested):

testing/e2e/manifests/clientServer.go
```diff
@@ -65,11 +65,11 @@ func ClientAndServer(namespace, name, nameserver, keyvaultURI, host, tlsHost str
 		},
 	}
 	clientDeployment.Spec.Template.Spec.Containers[0].ReadinessProbe = &corev1.Probe{
-		FailureThreshold:    1,
+		FailureThreshold:    6,
 		InitialDelaySeconds: 1,
-		PeriodSeconds:       1,
+		PeriodSeconds:       5,
```
jaiveerk, line 70: hm I mean realistically a single request shouldn't really be a source of any latency... would be curious why extending timeout here makes a difference

testing/e2e/suites/utils.go
```diff
@@ -36,14 +37,101 @@ func waitForAvailable(ctx context.Context, c client.Client, deployment appsv1.De
...
+	if err != nil {
+		return "", fmt.Errorf("building deployment selector: %w", err)
+	}
+
+	podList := &corev1.PodList{}
+	if err := c.List(ctx, podList, client.InNamespace(deployment.Namespace), client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		return "", fmt.Errorf("listing pods for deployment: %w", err)
+	}
+
+	podNames := make(map[string]struct{}, len(podList.Items))
```
jaiveerk, line 86: do we wanna tail pod logs here too? I feel like that's a diagnostic tool we've had in the past right

testing/e2e/manifests/embedded/gateway_client.golang
OliverMKing: what's the point of testing explicitly no such host? gets us nothing

depending on resolver/proxy timing it can surface as NXDOMAIN/no such host, timeout, or another dial/request failure. Keeping the check pinned to only no such host made the negative assertion sensitive to resolver implementation/timing rather than app-routing behavior

testing/e2e/manifests/clientServer.go
OliverMKing: The timeout change is not about application latency under normal success; it gives the kubelet probe enough budget for DNS resolution, TLS/dial, and transient resolver retries during propagation. The old FailureThreshold=1 / short timeout made a single slow transient mark the client unavailable immediately, which was enough to fail the e2e wait even if the route became healthy moments later

testing/e2e/manifests/embedded/gateway_client.golang
jaiveerk: I guess? I mean the point of the specific test was to tighten the scope of the negative test. If it's clear that no such host is not always the error that shows this specific kind of failure I'm fine with it, but I even wonder if we could do something else to prove that the dns record was never published.

While implementing this I considered hitting the zone directly but the client doesn't currently have any azure auth info, and this seemed like too niche of a case to bring it in.

Feel free to resolve if you'd like to keep this as is, I won't block on it either way.

testing/e2e/manifests/e2e.go
```diff
@@ -111,3 +114,23 @@ func E2e(image, loadableProvisionedJson string) []client.Object {
 
 	return ret
 }
+
+func e2eEnv() []corev1.EnvVar {
+	keys := []string{
+		"E2E_TEST_FILTER",
```
jaiveerk, line 120: where is this actually set

